PR #218, opened 2023-09-07
Assorted fixes and cleanup

### LINKED ISSUE #164: Add unit test for Laplace operator
In GitLab by @vinciguerra_a on May 3, 2023, 12:09

The Laplacian is the only field operator without a unit test.

### LINKED ISSUE #189: Identify and fix bug in master
In GitLab by @montan_v on Aug 10, 2023, 13:48

Encountered bug in PIC loops.
Applications and parameters failing:

1. ./PICnd 8 8 8 8, 8 nodes, 8 MPI ranks (1 MPI rank/node)

   Failure description:
   - After three iterations, domain scan fails to locate all particles.
2. ./LandauDamping 32 32 32 10000 , 4 nodes, 4 MPI ranks (1 MPI rank/node)

   Failure description:
   - Segmentation fault detected

Arc676:
An assortment of final changes to clean up the state of the master branch.

- Ensure that there is always a host space backend available by falling back to the stock FFTs if neither FFTW nor MKL are available. This ensures that the unit tests compile for "CUDA only" builds, which also include serial execution because Kokkos requires that at least one host space execution policy be defined.
- Update the unit test for `UniformCartesian` to use the new structure (overlooked in [!201](https://gitlab.psi.ch/OPAL/Libraries/ippl/-/merge_requests/201) since it doesn't use execution spaces at all).
- Add a unit test for the Laplacian; close #164.
- Fix field BCs unit test to actually use different execution spaces. Fixes various occurrences of incorrect execution space for range policy types.
- Remove the erroneous layout comparison introduced in [!192](https://gitlab.psi.ch/OPAL/Libraries/ippl/-/merge_requests/192) that prevented the periodic Poisson solver from updating it's temporary fields; close #189.
- Define in/equality operators for `NDIndex` (currently unused)


srikrrish (dismissed):

srikrrish (approved):

unit_tests/MultirankUtils.h
```diff
@@ -1,132 +0,0 @@
-//
-// Utility class for rank independent unit testing
```
matt-frey, line 2: Is this deleted on purpose?

unit_tests/Particle/ParticleBC.cpp
```diff
@@ -54,13 +54,10 @@ class ParticleBCTest<Parameters<T, ExecSpace, Rank<Dim>>> : public ::testing::Te
...
 
     void checkResult(const ippl::Vector<T, Dim>& expected) {
-        T tol = (std::is_same_v<T, double>) ? 1e-15 : 1e-7;
-
         Kokkos::deep_copy(mirror, bunch->R.getView());
-
         for (int i = 0; i < nParticles; ++i) {
             for (size_t j = 0; j < Dim; ++j) {
-                ASSERT_NEAR(expected[j], mirror(i)[j], tol);
+                ASSERT_NEAR(expected[j], mirror(i)[j], tolerance<T> / 10);
```
matt-frey, line 60: Why is the tolerance divided by 10? Shouldn't we then just provide another tolerance?

matt-frey (changes requested):

unit_tests/MultirankUtils.h
Arc676: Yes. It's been superseded by `TestUtils.h`. The more generic name reflects that it also covers testing with different precisions and execution spaces.

Testing with different dimensionalities is now covered by GTest parameterization instead of metaloops.

unit_tests/Particle/ParticleBC.cpp
Arc676: I asked about this in the group chat. We didn't reach a conclusion on a specific reason why this test had a lower tolerance than all the others. In the PR I just left it like this so that the expression reflects the nature of the test, but we can also make a new constant and call it `lower_tolerance` or similar.

matt-frey: Ok. I see.

matt-frey (approved):
Looks good!

PR #225, opened 2023-09-12
MPI communicator

matt-frey:
This PR
* changes the C-style array of type enumerator `{PARALLEL, SERIAL}` to a simple `std::array` of type boolean.
* adds a communicator to the field layout class
* adds the new communication interface (not yet complete)

This PR is created, although the new communication interface is not yet complete, because it is required for OPAL.


src/Communicate/Buffers.hpp
```diff
@@ -21,21 +21,20 @@
...
-            if (buffers[id]->getBufferSize() < size) {
+        template <typename MemorySpace, typename T>
+        Communicator::buffer_type<MemorySpace> Communicator::getBuffer(int id, size_type size,
+                                                                       double overallocation) {
+            auto& buffers = buffers_m.get<MemorySpace>();
+            size *= sizeof(T);
+            if (buffers.contains(id) && buffers[id]->getBufferSize() < size) {
                 buffers[id]->reallocBuffer(size);
             }
+            buffers[id] = std::make_shared<archive_type<MemorySpace>>(
```
srikrrish, line 34: I am a bit confused here.. This to me seems will always try to allocate the buffer with id even if we have enough size which we want to avoid because of performance in GPU. Previously we do this only if the id doesn't exist, or did I misunderstand something here?

src/Communicate/Communicator.h
```diff
@@ -0,0 +1,219 @@
...
+            template <typename MemorySpace = Kokkos::DefaultExecutionSpace::memory_space>
+            void deleteBuffer(int id) {
+                buffers_m.get<MemorySpace>().erase(id);
+            }
+
+            void deleteAllBuffers();
+
+            const MPI_Comm& getCommunicator() const noexcept { return *comm_m; }
+
+            //         void setCommunicator(const MPI_Comm& comm) noexcept { comm_m.reset(comm); }
```
srikrrish, line 158: Why is this commented? Should we just remove?

srikrrish (changes requested):
I would have expected that the things in FFT class should also be changed as they now have to use the layout's communicator instead of the global communicator. Is it already there? I searched for it but unable to find.

matt-frey:
I have not yet touched the FFTs. Note that the default communicator for the field layout is MPI_COMM_WORLD.

src/Communicate/Collectives.hpp
```diff
@@ -0,0 +1,63 @@
...
+        void Communicator::reduce(const T* input, T* output, int count, Op op, int root) {
+            MPI_Datatype type = get_mpi_datatype<T>(*input);
+
+            MPI_Op mpiOp = get_mpi_op<Op>(op);
+
+            MPI_Reduce(const_cast<T*>(input), output, count, type, mpiOp, root, *comm_m);
+        }
+
+        template <typename T, class Op>
+        void Communicator::reduce(const T& input, T& output, int count, Op op, int root) {
```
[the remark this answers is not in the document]
matt-frey: Yes, I thought it would be nice have a reference version for scalars, but will think about this.

srikrrish:
I thought Alex also changed the communicators in FFTs so that they use the layout's communicator. Anyway, I think it might be better to merge the branches #218 and #217 create a release and a tag and then merge this branch. Because I have a feeling this may also potentially break a few things and right now we don't have time to test all of the codes etc. What do you think?

matt-frey:
Sure. Alex did this in his fork only. I was not yet able to apply the new MPI interface to all components of Ippl. Still WIP.

srikrrish:
That's fine..

src/Communicate/Buffers.hpp
Arc676: This is a mistake caused by a merge conflict. The buffer should be returned after it's reallocated when the buffer already exists.

Arc676: No I think it was one of my commits that lost the `return` and I just didn't notice at the time.

matt-frey: Ok.

srikrrish (approved):

src/Communicate/Buffers.hpp
srikrrish: It still seems to me that this problem was not fixed. Or am I wrong?

srikrrish (changes requested):

srikrrish (approved):

aaadelmann (approved):

PR #230, opened 2023-09-26
Solver design refactoring

s-mayani:
We refactor the solver according to the proposed design in the attached slides.
[solver_design.pdf](https://github.com/IPPL-framework/ippl/files/12729799/solver_design.pdf)


src/Solver/FFTPeriodicPoissonSolver.h
```diff
@@ -1,6 +1,6 @@
 //
 // Class FFTPeriodicPoissonSolver
-//   Solves periodic electrostatics problems using Fourier transforms
+//   Solves the periodic Poisson problem using Fourier transforms
```
aaadelmann, line 3: Please add a reference to the original algorithm.

src/Solver/Maxwell.h
```diff
@@ -0,0 +1,85 @@
+//
+// Class Maxwell
+//   Base class for solvers for Maxwell's equations
+//
```
aaadelmann, line 4: Also here we should mention the original paper Mithra !?

src/Solver/Maxwell.h
```diff
@@ -0,0 +1,85 @@
...
+            setSources(charge, current);
+            setEMFields(E, B);
+        }
+
+        /*!
+         * Set the problem RHS (charge & current densities)
+         * @param charge Reference to rho field
+         * @param current Reference to J field
+         */
+        virtual void setSources(SourceField& charge, VectorSourceField_t& current) {
```
aaadelmann, line 52: is there a reason why you have both set functions virtual?

src/Solver/Maxwell.h
```diff
@@ -0,0 +1,85 @@
...
+        /*!
+         * Solve the electromagnetic problem (Maxwell's eqs)
+         */
+        virtual void solve() = 0;
+
+        virtual ~Maxwell() {}
+
+    protected:
+        // fields containing reference to charge and current
+        SourceField* rhoN_mp       = nullptr;
```
aaadelmann, line 76: why not assigning the nullprt in the constructor?

aaadelmann (changes requested):

src/Solver/Maxwell.h
s-mayani: I think I will change this to have setSources as virtual and setEMFields not virtual, such that the specific initialize() of the subclasses can be called in setSources (just like it is in Poisson.h). Thanks for pointing it out!

src/Solver/Maxwell.h
s-mayani: I would cite Mithra in the class FDTDSolver, not in the base Maxwell class (as the FEM solver will also inherit from this).

aaadelmann (approved):

src/Solver/Maxwell.h
matt-frey: Consider smart pointers.

matt-frey (changes requested):

src/Solver/FFTPoissonSolver.h
```diff
@@ -70,15 +70,15 @@ namespace ippl {
     }  // namespace detail
 
     template <typename FieldLHS, typename FieldRHS>
-    class FFTPoissonSolver : public Electrostatics<FieldLHS, FieldRHS> {
+    class FFTPoissonSolver : public Poisson<FieldLHS, FieldRHS> {
```
srikrrish, line 73: May not be needed in the current PR but would it make sense to rename this as "FFTOpenPoissonSolver" as this works only with open BCs and the naming is then also consistent with FFTPeriodicPoissonSolver. I agree that as Andreas said it is not very nice to have the BCs name also in the solver class but in my opinion these are very specific solvers and they are also fundamentally different as we already discussed (Integral Vs Differential form).

srikrrish (commented):
I like that we now separate into LinearSolvers and equation specific solvers. But is then Solver and LinearSolvers sort of confusing? Ideally instead of `Solver` directory there should be PoissonSolvers, MaxwellSolvers and may be in future we have NavierStokesSolvers etc. Then the files corresponding to them can be moved into the appropriate directories I think.

src/Solver/FFTPoissonSolver.h
s-mayani: I think conceptually it makes sense, and I can do it in this PR too, if everyone agrees.

src/Solver/Maxwell.h
s-mayani: In Poisson.h, we use normal pointers for rhs_mp and lhs_mp, and then for example in the FFTPoissonSolver the mesh and layout of the rhs_mp is also a normal pointer, but all other pointers are smart. I do not know why this is like this, but if I change to smart pointers here, it would make sense to change everywhere in the solvers. Should I do that?

s-mayani: Actually after reading a bit more on it, I think we should leave all the solver class members which will be pointers to fields coming from the outside (e.g. rhs and lhs) as normal pointers, not smart, since otherwise we need to allocate using `new` to initialize rhs and lhs. What do you think?

srikrrish (approved):

matt-frey (approved):

PR #232, opened 2023-11-02
127 nearest neighbor loop

vmontanaro:


src/Particle/ParticleSpatialLayout.h
```diff
@@ -73,11 +75,18 @@ namespace ippl {
         //! The RegionLayout which determines where our particles go.
         RegionLayout_t rlayout_m;
 
-        using region_type = typename RegionLayout_t::view_type::value_type;
+        //! The FieldLayout containing information on nearest neighbors
+        FieldLayout_t flayout_m;
+
+	    using region_view_type  = typename RegionLayout_t::view_type;
```
aaadelmann, line 81: add a comment what is the purpose of these types

src/Particle/ParticleSpatialLayout.h
```diff
@@ -73,11 +75,18 @@ namespace ippl {
...
+        using region_type 	= typename region_view_type::value_type;
+	    using neighbor_list = typename FieldLayout_t::neighbor_list;
 
         template <size_t... Idx>
         KOKKOS_INLINE_FUNCTION constexpr static bool positionInRegion(
             const std::index_sequence<Idx...>&, const vector_type& pos, const region_type& region);
+        
+        size_type getNeighborSize(const neighbor_list& neighbors) const;
```
aaadelmann, line 89: add a comment what is the purpose of the function. use Doxygen. This is valid for other functions (and types)

src/Particle/ParticleSpatialLayout.hpp
```diff
@@ -27,10 +27,32 @@
 
 namespace ippl {
 
+    struct increment_type {
```
aaadelmann, line 30: same comment than above

src/Particle/ParticleSpatialLayout.hpp
```diff
@@ -162,34 +184,139 @@ namespace ippl {
...
 
         const auto is = std::make_index_sequence<Dim>{};
 
-        size_type invalidCount = 0;
-        Kokkos::parallel_reduce(
+        const neighbor_list& neighbors = flayout_m.getNeighbors();
+
+        // container of particles that travelled more than one cell
+        locate_type notFoundIds("Not found", size_type(0.1 * pc.getLocalNum()));
+        // Now: dimension hard-coded, for future implementations maybe make it as a run parameter.
```
aaadelmann, line 216: please add this comment as a Doxygen \todo

src/Particle/ParticleSpatialLayout.hpp
```diff
@@ -162,34 +184,139 @@ namespace ippl {
...
+        for (const auto& componentNeighbors : neighbors) {
+            for (size_t j = 0; j < componentNeighbors.size(); ++j) {
+                neighbors_mirror(k) = componentNeighbors[j];
+                k++;
+            }
+        }
+
+        Kokkos::deep_copy(neighbors_view, neighbors_mirror);
+
+        /*Begin Kokkos loop:
```
aaadelmann, line 240: Please make this Doxygen like

aaadelmann (commented):
In general pleas work a bit on the code documentaion.

src/Particle/ParticleSpatialLayout.hpp
```diff
@@ -162,34 +184,139 @@ namespace ippl {
...
-                bool xyz_bool = positionInRegion(is, positions(i), Regions(j));
+            Kokkos::RangePolicy<size_t>(0, ranks.extent(0)),
+            KOKKOS_LAMBDA(const size_type i, increment_type& val, const bool final) {
+                // Step 1
+                bool xyz_bool = false;
+                bool increment[2];
+
+                xyz_bool = positionInRegion(is, positions(i), Regions(myRank));
+
                 if (xyz_bool) {
```
srikrrish, line 256: I think all these `if`s can be avoided by doing something similar to what we do in other places in IPPL
`ranks(i) = xyz_bool * rank;
invalid(i) = !xyz_bool;
found(i) = xyz_bool;
`
This will put `0` as the rank for the particles which are not found in the current rank. But it is ok since that will be replaced by the correct ranks in the steps 2 and 4. Similar to above the `if`s in steps 2 and 4 can also be avoided.

srikrrish (changes requested):

src/Particle/ParticleSpatialLayout.hpp
vmontanaro: The only doubt i have on removing the `if` is how can i handle the break from the loop

vmontanaro: But the other ones can be removed, you're right

src/Particle/ParticleSpatialLayout.h
vmontanaro: See latest commit (for all comments regarding doxygen)

aaadelmann (approved):

src/Particle/ParticleSpatialLayout.hpp
srikrrish: Why do you need to break from that loop? Even if you don't break you are going to test for the other field neighbors (which is not needed) but since this is small (O(10)s only) and GPUs especially don't care about it in my opinion we don't need that break.

vmontanaro: My question is:

Say `xyz_found` is true for the first neighbor. The function assigns `invalid(i) = xyz_found`, then at the next cycle you get `xyz_found = false.` If you reassign `invalid(i) = xyz_found`, the previous true valued will  be overwritten...

aaadelmann: > @aaadelmann I usually check what comments are resolved/unresolved and approve based on that. So in future it would be of great help if you could check carefully before marking discussions as resolved. Because I accidentally skipped this one which in my opinion is still unresolved.

Sorry I think this is because of my lack of understanding of the github workflow. However I see that this is not approved yet so I guess I do not understand the issue ... because of my lack of understanding of the github workflow. Lets maybe discuss this on a Zoom so I better get the ideas.

srikrrish: @aaadelmann No problem I am also not 100 percent sure with the github workflow and trying to figure out things which were in the gitlab workflow. Yes in this case it is fine as I somehow noticed Veronica's message before and searched through the resolved discussions to find this one. But I usually don't do that and just look at the PR again and see what things are still unresolved (assuming the things which are resolved are ok). Anyway as you mentioned a Zoom may be better also with @matt-frey as he is also more experienced with github workflow.

srikrrish: @vmontanaro Yeah didn't think about it before. How about `invalid(i) += xyz_found` would that create a race condition? Since the Kokkos loop is over the particles multiple threads should not write but I am not entirely sure..

vmontanaro: I could try with that one, it should not give problems.

vmontanaro: > @vmontanaro Yeah didn't think about it before. How about `invalid(i) += xyz_found` would that create a race condition? Since the Kokkos loop is over the particles multiple threads should not write but I am not entirely sure..

This gives a `Kokkos::Impl::SharedAllocationRecord 'Found' failed decrement count = 0`.

vmontanaro: > > @vmontanaro Yeah didn't think about it before. How about `invalid(i) += xyz_found` would that create a race condition? Since the Kokkos loop is over the particles multiple threads should not write but I am not entirely sure..
> 
> @srikrrish This gives a `Kokkos::Impl::SharedAllocationRecord 'Found' failed decrement count = 0`.

Fixed in last commit.

srikrrish: Yeah this seems to be good too..

srikrrish (approved):

PR #233, opened 2023-11-13
Update unit test framework

felixschurk:
This pull request updates to work with the way how currently GoogleTest is added to projects following the setup for CMake [0].
This approach works without the need to load the `gtest` module or build it manually, allowing the unit tests to be always run, everything is handled automatically.

Further as then the unit tests are also registered with the cmake build system, and therefore also with `ctest` one can start all the unit tests over
```sh
ctest --test-dir <build-dir>/unit_tests
```
Having the option to work with `ctest` then also allows to let the unit tests run in parallel or incorporate them more easily into build scripts, more information see ctest documentation [1], possibly the command looks like
```sh
ctest --test-dir <build-dir>/unit_tests -j 6 --progress 
```

The `FindGTest.cmake` is therefore no longer needed and thus also deleted.
As the `FetchContent` module in CMake spits out a warning, starting from version 3.24 policy CMP0135 is set to `NEW` [2].

[0]: https://google.github.io/googletest/quickstart-cmake.html
[1]: https://cmake.org/cmake/help/latest/manual/ctest.1.html
[2]: https://cmake.org/cmake/help/latest/policy/CMP0135.html#policy:CMP0135


aaadelmann:
Hi Felix this is a great feature. We should also have an entry in our IPPL documentation.  @srikrrish any idea where this should go?

srikrrish:
> Hi Felix this is a great feature. We should also have an entry in our IPPL documentation. @srikrrish any idea where this should go?

There is a page about unit tests in the documentation https://github.com/IPPL-framework/ippl/blob/master/UNIT_TESTS.md but it is more about creating new unit tests rather than testing with it. I would suggest this could be part of build scripts and the associated documentation https://github.com/IPPL-framework/ippl-build-scripts#readme so that once IPPL is built the ctest can run the unit tests and check if the installation is fine.

CMakeLists.txt
```diff
@@ -79,11 +79,25 @@ if (ENABLE_ALPINE)
     add_subdirectory (alpine)
 endif ()
 
-option (ENABLE_UNIT_TESTS "Enable unit tests" OFF)
+option (ENABLE_UNIT_TESTS "Enable unit tests" ON)
```
srikrrish, line 82: All the `ENABLE` options in IPPL are by default `OFF` as we follow the "opt in" way rather than  "opt out".
Although it may make sense for unit tests I am not sure if we want to change this behavior as it is a bit inconsistent with the other `ENABLE` options.

CMakeLists.txt
```diff
@@ -79,11 +79,25 @@ if (ENABLE_ALPINE)
...
 if (ENABLE_UNIT_TESTS)
-    find_package (GTest REQUIRED HINT $ENV{GTEST_PREFIX})
-    add_subdirectory (unit_tests)
-    #add_subdirectory(unit_tests_single_precision)
+  include(FetchContent)
```
srikrrish, line 84: Could you run the code formatting? The indent seems to be off compared to the other lines (see the above `if`) in this file.

srikrrish (changes requested):

CMakeLists.txt
felixschurk: https://github.com/IPPL-framework/ippl/pull/233/commits/97c7b60d5e85396a3a62561c8706acc4dbc876a4

Resolved in this commit, however the automatic code formatting of `clang-format` is only working for `*.h, *.hpp` and `*.cpp` files.

CMakeLists.txt
felixschurk: https://github.com/IPPL-framework/ippl/pull/233/commits/7148da36893711cbca529b391806e5f0d574e238

Taken care of :)

srikrrish (approved):

Arc676:
Seems good; although as a side note, this means that any CI jobs will always download and rebuild `gtest`, which is more node time if we use Merlin for GPU testing as part of the pipeline (assuming we use a clean build).

Arc676 (approved):

felixschurk:
@Arc676 
Yes with the always build that is a valid point which I did not had yet in mind. 

I would say if that is really an issue (in terms of large increase in compile time etc.), then it would make sense to extend the `CMakeLists.txt` to first check if a local `gtest` would be available and otherwise pull down the repository.

Maybe to have it as a reference if this is needed in the future, the ADIOS2 library has this approach available.

https://github.com/ornladios/ADIOS2/blob/a8e02fe52b1760877bf800c7065b9bf5d46c4c16/thirdparty/CMakeLists.txt#L23-L49

CMakeLists.txt
```diff
@@ -81,9 +81,23 @@ endif ()
...
-    find_package (GTest REQUIRED HINT $ENV{GTEST_PREFIX})
+    include(FetchContent)
+    include(GoogleTest)
+
+    enable_testing()
+
+    FetchContent_Declare(
+        googletest
+        # currently v1.14
+        URL https://github.com/google/googletest/archive/f8d7d77c06936315286eb55f8de22cd23c188571.zip
```
matt-frey, line 92: Wouldn't it be better to use a link with the version, i.e. `https://github.com/google/googletest/archive/refs/tags/v1.14.0.zip`? One could then also provide the version as an optional argument to CMake.

matt-frey (commented):
Looks good to me. Regarding the comment about CI. It may also be possible that all dependencies have to be built from scratch anyway, every time they are run. It depends on the system we are on. On PSI systems, these modules are available, but on others maybe not.

CMakeLists.txt
felixschurk: Very valid point, this is more expressive and directly clear to read.
Added in https://github.com/IPPL-framework/ippl/pull/233/commits/c8b366fc89229bf92821c0ccae03b0072ea0f220.

matt-frey (approved):

PR #240, opened 2023-11-27
Pic manager

mohsensadr:
Updating the pic-manager in the dev-3.2.0 including:
-making member variables of defined classes private.
-adding get/set functions.
-removing field solver from template parameter of pic-manager.
-revising the MPI calls according to dev-3.2.0.


aaadelmann (approved):

aaadelmann (approved):

alpine/FieldContainer.hpp
```diff
@@ -4,42 +4,64 @@
...
+        ippl::e_dim_tag decomp_m[Dim];
+        Vector_t<double, Dim> hr_m;
+        Vector_t<double, Dim> rmin_m;
+        Vector_t<double, Dim> rmax_m;
+        std::shared_ptr<Mesh_t<Dim>> mesh_m;
+        std::shared_ptr<FieldLayout_t<Dim>> fl_m;
+
+    public:
+        VField_t<T, Dim>& getE() { return E_m; }
+        void setE(VField_t<T, Dim>& E) { E_m = E; }
```
aaadelmann: we should inline  all get/set functions

aaadelmann (approved):

alpine/FieldContainer.hpp
mohsensadr: Is this what you mean? see [347eeef](https://github.com/IPPL-framework/ippl/pull/240/commits/347eeefc541f2888f6a19c2130d77cb2a4a1d894)

```
    inline VField_t<T, Dim>& getE() { return E_m; }
    inline void setE(VField_t<T, Dim>& E) { E_m = E; }
```

aaadelmann: perfect

mohsensadr:
@srikrrish @matt-frey 
Hi, please have a look and let me know if you have any comments/feedback.

srikrrish (commented):
If I am not wrong BumponTailInstability still has the old structure. I think it is better to change that as well to be conformant with the other two mini-apps.

mohsensadr:
> If I am not wrong BumponTailInstability still has the old structure. I think it is better to change that as well to be conformant with the other two mini-apps.

Added.

alpine/BumponTailInstability.cpp
```diff
@@ -15,12 +15,21 @@
...
+//
+// This file is part of IPPL.
+//
+// IPPL is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// You should have received a copy of the GNU General Public License
+// along with IPPL. If not, see <https://www.gnu.org/licenses/>.
```
matt-frey, line 31: Please remove this.

alpine/FieldContainer.hpp
```diff
@@ -29,30 +29,29 @@ class FieldContainer{
     std::shared_ptr<FieldLayout_t<Dim>> fl_m;
 
 public:
-    VField_t<T, Dim>& getE() { return E_m; }
-    void setE(VField_t<T, Dim>& E) { E_m = E; }
+    inline VField_t<T, Dim>& getE() { return E_m; }
+    inline void setE(VField_t<T, Dim>& E) { E_m = E; }
```
matt-frey, line 33: Note that all these inlines are superfluous. All functions entirely defined in the class are already inlined. Read for example https://en.cppreference.com/w/cpp/language/inline

alpine/LandauDampingManager.h
```diff
@@ -418,7 +410,7 @@ class LandauDampingManager
             Inform csvout(NULL, fname.str().c_str(), Inform::APPEND);
             csvout.precision(16);
             csvout.setf(std::ios::scientific, std::ios::floatfield);
-            if (time_m == 0.0) {
+            if ( fabs(time_m) < 1e-14 ) {
```
matt-frey, line 413: Please add the namespace for `fabs`.

alpine/LoadBalancer.hpp
```diff
@@ -22,20 +22,20 @@ class LoadBalancer{
 
         ~LoadBalancer() {  }
 
-        double getLoadBalanceThreshold() const { return loadbalancethreshold_m; }
-        void setLoadBalanceThreshold(double threshold) { loadbalancethreshold_m = threshold; }
+        inline double getLoadBalanceThreshold() const { return loadbalancethreshold_m; }
+        inline void setLoadBalanceThreshold(double threshold) { loadbalancethreshold_m = threshold; }
```
matt-frey, line 26: Same comment regarding the inline keyword. It is superfluous.

src/Manager/FieldSolverBase.h
```diff
@@ -62,10 +62,11 @@ using Solver_t = VariantFromConditionalTypes<CGSolver_t<T, Dim>, FFTSolver_t<T,
 namespace ippl {
   template <typename T, unsigned Dim>
   class FieldSolverBase {
```
matt-frey, line 64: I do not yet see the purpose of this abstract base class.

alpine/BumponTailInstability.cpp
```diff
@@ -15,12 +15,21 @@
 //     ovfactor = Over-allocation factor for the buffers used in the communication. Typical
 //                values are 1.0, 2.0. Value 1.0 means no over-allocation.
 //     Example:
-//     srun ./BumponTailInstability 128 128 128 10000 10 FFT 0.01 --overallocate 2.0 --info 10
-//     Change the TestName to TwoStreamInstability or BumponTailInstability
-//     in order to simulate the Two stream instability or bump on tail instability
-//     cases
+//     srun ./BumponTailInstability 128 128 128 10000 10 FFT 0.01 LeapFrog --overallocate 2.0 --info 10
+//
+// Copyright (c) 2023, Paul Scherrer Institut, Villigen PSI, Switzerland
```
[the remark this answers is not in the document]
mohsensadr: done in 815d239.

alpine/BumponTailInstability.cpp
mohsensadr: Done in 815d239.

alpine/LandauDampingManager.h
mohsensadr: Added std as the namespace in 9926690.

src/Manager/FieldSolverBase.h
mohsensadr: I wanted to remove FieldSolver from the list of template arguments of PicManager. We need a type here to make shared pointer fsolver_m out of it. However, the FieldSolver class is defined in alpine directory. I find two solutions:

1. We can create a base class for the field solver in src/Manager/. (FieldSolverBase) with very basic member variables. Then, use it to create fsolver_m pointer in pic-manager, and complete/extend it in the FieldSolver class in alpine/ directory by inheritance.

2. In src/Manager/PicManager.h we can include the field solver class from alpine/FieldSolver.hpp and use the class FieldSolver as a type directly to create fsolver_m pointer.

I implemented the 1st solution, which doesn't require including code from alpine/. directory. 
Does it make sense?

alpine/BumponTailInstabilityManager.h
```diff
@@ -0,0 +1,579 @@
...
+#include "Random/Randn.h"
+
+using view_type = typename ippl::detail::ViewType<ippl::Vector<double, Dim>, 1>::view_type;
+
+const char* TestName = "BumponTailInstability";
+
+constexpr bool EnablePhaseDump = false;
+
+// define functions used in sampling particles
+struct CustomDistributionFunctions {
```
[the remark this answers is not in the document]
mohsensadr: I'm not sure what you mean. The struct "CustomDistributionFunctions" is gonna be used as the template argument to make an instant of Distribution class

```
using DistR_t = ippl::random::Distribution<double, Dim, 2 * Dim, CustomDistributionFunctions>;
...
DistR_t distR(parR);
```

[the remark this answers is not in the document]
srikrrish: Ok somehow this is very confusing to me.. In my opinion if a mini-app requires custom distribution it should derive from the base class Distribution and just implement their own PDF, CDF etc. The base class Distribution would then be an abstract class with PDF, CDF etc. as pure virtual functions. This has clear hierarchy in my opinion. Or what problems did you face if you implement in this way?

srikrrish:
I just saw from the PenningTrapManager that this PR does not seem to conform to coding style requirements. That is `_m` for class member variables etc. It also sometimes seem to put `_m` for local variables inside member functions which should not be the case. Please fix it. I am unable to comment on the exact locations as this PR only shows partial content.

alpine/BumponTailInstabilityManager.h
```diff
@@ -0,0 +1,579 @@
...
+        Inform m("Initialize Particles");
+
+        using DistR_t = ippl::random::Distribution<double, Dim, 2 * Dim, CustomDistributionFunctions>;
+        double* parR  = new double[2 * Dim];
+        parR[0]       = delta;
+        parR[1]       = kw[0];
+        parR[2]       = delta;
+        parR[3]       = kw[1];
+        parR[4]       = delta;
+        parR[5]       = kw[2];
```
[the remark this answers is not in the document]
mohsensadr: I enabled CG and tested the mini-apps. Also, I tested 1D and 2D cases of Bump on Tail and Landau Damping. It seems to reproduce the results for serial and multiple GPUs.

mohsensadr:
@srikrrish I updated the name of member variables 3843ce8
I guess the updated version should be ok unless I missed something.

alpine/FieldContainer.hpp
mohsensadr: @aaadelmann shall I remove `inline` from all the get/set functions?  As @matt-frey pointed out, it seems to be redundant.

alpine/BumponTailInstabilityManager.h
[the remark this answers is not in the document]
srikrrish: Ok. What about Sonali's open boundary solver and P3M solver? If I am not wrong they were only for Penning trap I think. Could you please test that too? Also it would be good to test some of these cases on multiple CPU only builds too.

alpine/LandauDampingManager.h
```diff
@@ -116,116 +112,118 @@ class LandauDampingManager
...
+            parR[i * 2   ]  = alpha_m;
+            parR[i * 2 + 1] = kw_m[i];
+        }
         DistR_t distR(parR);
 
-        Vector_t<double, Dim> kw_m     = this->kw;
-        Vector_t<double, Dim> hr_m     = hr;
-        Vector_t<double, Dim> origin_m = origin;
-        if ((this->loadbalancethreshold_m != 1.0) && (ippl::Comm->size() > 1)) {
+        Vector_t<double, Dim> kw     = kw_m;
```
srikrrish, line 208: I see in so many places you define new temporary variables. Why are they needed? and why can't you directly use here `kw_m`? Note these things increase the lines of code (which we want to minimize) and gives an impression of a complex code even though the code itself might be simple only.

srikrrish: Is it because of `KOKKOS_LAMBDA` vs `KOKKOS_CLASS_LAMBDA`? In this case we may then want to not specify so many of them in the class and define some of them just locally rather than having it in the class as well as defining a temporary variable.

alpine/LandauDampingManager.h
```diff
@@ -116,116 +112,118 @@ class LandauDampingManager
...
+        dt_m     = std::min(.05, 0.5 * *std::min_element(hr_m.begin(), hr_m.end()));
+        it_m     = 0;
+        time_m   = 0.0;
 
         m << "Discretization:" << endl
-          << "nt " << this->nt << " Np= " << this->totalP << " grid = " << this->nr << endl;
+          << "nt " << nt_m << " Np= " << totalP_m << " grid = " << nr_m << endl;
 
-        std::shared_ptr<Mesh_t<Dim>> mesh = std::make_shared<Mesh_t<Dim>>(this->domain, this->hr, this->origin);
+        std::shared_ptr<Mesh_t<Dim>> mesh = std::make_shared<Mesh_t<Dim>>(domain_m, hr_m, origin_m);
```
srikrrish, line 146: Why do you need them as pointers unlike the previous implementation in master? At least to me it makes the code looks more complex with more characters.

alpine/LandauDampingManager.h
```diff
@@ -116,116 +112,118 @@ class LandauDampingManager
...
 
-        this->isAllPeriodic = true;
+        isAllPeriodic_m = true;
 
-        std::shared_ptr<FieldLayout_t<Dim>> FL = std::make_shared<FieldLayout_t<Dim>>(MPI_COMM_WORLD, this->domain, this->decomp, this->isAllPeriodic);
+        std::shared_ptr<FieldLayout_t<Dim>> FL = std::make_shared<FieldLayout_t<Dim>>(MPI_COMM_WORLD, domain_m, decomp_m, isAllPeriodic_m);
```
srikrrish, line 150: Same here. Why do you need them as a pointer?

srikrrish (changes requested):

alpine/LandauDampingManager.h
```diff
@@ -324,59 +349,64 @@ class LandauDampingManager
 
     void gatherCIC() {
         using Base                        = ippl::ParticleBase<ippl::ParticleSpatialLayout<T, Dim>>;
-        Base::particle_position_type *E_p = &this->pcontainer_m->getE();
-        Base::particle_position_type *R_m = &this->pcontainer_m->R;
-        VField_t<T, Dim> *E_f             = &this->fcontainer_m->getE();
-        gather(*E_p, *E_f, *R_m);
+        Base::particle_position_type *Ep = &pcontainer_m->E;
```
srikrrish, line 352: Here why can't you directly pass `pcontainer_m->E` and need to define these temporaries?

alpine/PenningTrapManager.h
```diff
@@ -229,46 +225,52 @@ class PenningTrapManager
...
-        this->pcontainer_m->getQ() = this->Q / this->totalP;
+        IpplTimings::stopTimer(particleCreation);
+
+        pcontainer_m->q = Q_m / totalP_m;
         m << "particles created and initial conditions assigned " << endl;
     }
 
     void advance() override {
-        if (this->step_method == "LeapFrog") {
+        if (stepMethod_m == "LeapFrog") {
```
srikrrish, line 273: Should have an assert here for checking if the stepper is not at all specified or if anything other than `LeapFrog` is specified.

srikrrish (commented):
I think we should move UniformPlasmaTest.cpp, ChargedParticles.hpp, LandauDampingMixedExec.cpp and LandauDampingMixedPrecision.cpp to a separate new directory as having them here makes it confusing.

srikrrish:
@Arc676 I requested a review from you as you worked on redesigning the mini-apps interfaces quite extensively. But feel free to skip it if you don't have time which is perfectly fine. Note since some parts of this are already merged if you want to have a full overview unfortunately you have to directly look at the files and also may not be able to directly comment on the lines which you would like to.

alpine/LandauDampingManager.h
mohsensadr: Yes, I use local variables so that the KOKKOS_LAMBDA doesn't copy all the member variables of the class in the loop. I'll try to reduce local variables as much as possible in the next commit

srikrrish: Yes.. also if that is not possible try reducing the number of variables in the class (if they are not used everywhere) and just have them only defined as local variables

alpine/LandauDampingManager.h
mohsensadr: mesh and FL are created inside the method pre_run and then are used to set the corresponding member variables/pointers in the field container. If I allocate them statically without using pointers, I get invalid memory access outside the scope in other parts of the code.

alpine/LandauDampingManager.h
mohsensadr: see above

src/Manager/FieldSolverBase.h
matt-frey: Yes, It is a good idea to remove the field solver from the list of template arguments. No, the second solution is definitely not a good idea. I agree with your choice.

alpine/LoadBalancer.hpp
```diff
@@ -12,30 +12,34 @@ class LoadBalancer{
...
         std::shared_ptr<ParticleContainer<T, Dim>> pc_m;
         std::shared_ptr<FieldSolver_t> fs_m;
         unsigned int loadbalancefreq_m;
         ORB<T, Dim> orb;
     public:
         LoadBalancer(double lbs, std::shared_ptr<FieldContainer<T,Dim>> &fc, std::shared_ptr<ParticleContainer<T, Dim>> &pc, std::shared_ptr<FieldSolver_t> &fs)
-           :loadbalancethreshold_m(lbs), rho_m(&fc->getRho()), E_m(&fc->getE()), pc_m(pc), fs_m(fs) {}
+           :loadbalancethreshold_m(lbs), rho_m(&fc->getRho()), E_m(&fc->getE()), phi_m(&fc->getPhi()), pc_m(pc), fs_m(fs) {}
 
         ~LoadBalancer() {  }
```
Arc676, line 24: This entire class could also be part of a common ALPINE-level manager superclass. Depending on how the PIC manager is intended to work, it could also be moved all the way up to the generic PIC manager class and accessed via generic interface. Aside from the behavior specific to the uniform plasma test, the partitioning code itself is not application-specific.

alpine/ParticleContainer.hpp
```diff
@@ -9,10 +9,11 @@ template <typename T, unsigned Dim = 3>
 class ParticleContainer : public ippl::ParticleBase<ippl::ParticleSpatialLayout<T, Dim>>{
```
Arc676, line 9: If the particle container class derives from `ParticleBase`, why is the particle container a template parameter of the PIC manager? Why isn't the member just a pointer to `ParticleBase`?

alpine/BumponTailInstabilityManager.h
```diff
@@ -0,0 +1,613 @@
+#ifndef IPPL_BUMPON_TAIL_INSTABILITY_MANAGER_H
```
Arc676, line 1: Missing licensing information at top of file.

Arc676 (changes requested):
Some of the code (not necessarily introduced in this PR) is not properly formatted. (I didn't fully read through all the changes in the PR; the changes to the existing managers seem to be interface changes so I just skimmed through those.)

alpine/LandauDampingManager.h
srikrrish: In this case can't they be members of `FieldContainer`? After all they are field related entities..

alpine/BumponTailInstabilityManager.h
```diff
@@ -0,0 +1,613 @@
...
+
+    void grid2par() override { gatherCIC(); }
+
+    void gatherCIC() {
+        using Base                       = ippl::ParticleBase<ippl::ParticleSpatialLayout<T, Dim>>;
+        Base::particle_position_type *Ep = &pcontainer_m->E;
+        Base::particle_position_type *R  = &pcontainer_m->R;
+        VField_t<T, Dim> *Ef             = &fcontainer_m->getE();
+        gather(*Ep, *Ef, *R);
+    }
```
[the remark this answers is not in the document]
srikrrish: You mean an ALPINE-level class which is derived from`PicManager` and the individual managers like `LandauDampingManager` are then derived from it?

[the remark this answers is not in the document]
srikrrish: Yes I agree we should avoid code repetition as much as possible..

alpine/BumponTailInstabilityManager.h
srikrrish: @Arc676 After switching to the BSD-3 license we stopped putting licensing info at top of every file. Rather there is only one file at the root level of the repo. I assume you were asking about that only.

alpine/BumponTailInstabilityManager.h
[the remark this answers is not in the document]
Arc676: Yes, exactly. `ChargedParticles` contained all the common elements (and a few things it arguably shouldn't have contained) but otherwise we should employ the same strategy; a common manager class for all the ALPINE apps can hold the common elements.

alpine/ParticleContainer.hpp
mohsensadr: Actually, I tried it before. The issue is that then `pcontainer_m` of `PicManager` in src/Manager/PicManager.h would be a pointer of type `ippl::ParticleBase<ippl::ParticleSpatialLayout<T, Dim>` which doesn't have P,q,E as public variables. Then, the instance of ParticleContainer in the alpine application cannot be used to set `pcontainer_m` of `PicManager`, since the public members (P,q,E) are not declared in the base class.


Maybe there's a way to do it nicely, but I when I tried,  got errors like this

```
error: ‘using element_type = class ippl::ParticleBase<ippl::ParticleSpatialLayout<double, 3, ippl::UniformCartesian<double, 3> > >’ {aka ‘class ippl::ParticleBase<ippl::ParticleSpatialLayout<double, 3, ippl::UniformCartesian<double, 3> > >’} has no member named ‘P’
  266 |         view_type* P = &(pcontainer_m->P.getView());
      |                                        ^
/psi/home/sadr_m/Codes/ippl_fresh/ippl/alpine/LandauDampingManager.h:280:23: error: ‘using element_type = class ippl::ParticleBase<ippl::ParticleSpatialLayout<double, 3, ippl::UniformCartesian<double, 3> > >’ {aka ‘class ippl::ParticleBase<ippl::ParticleSpatialLayout<double, 3, ippl::UniformCartesian<double, 3> > >’} has no member named ‘q’
  280 |         pcontainer_m->q = Q_m/totalP;
      |                       ^
/psi/home/sadr_m/Codes/ippl_fresh/ippl/alpine/LandauDampingManager.h: In member function ‘void LandauDampingManager::LeapFrogStep()’:
/psi/home/sadr_m/Codes/ippl_fresh/ippl/alpine/LandauDampingManager.h:303:51: error: conversion from ‘shared_ptr<ippl::ParticleBase<ippl::ParticleSpatialLayout<double, 3, ippl::UniformCartesian<double, 3> > >>’ to non-scalar type ‘shared_ptr<ParticleContainer<double, 3>>’ requested
  303 |         std::shared_ptr<ParticleContainer_t> pc = pcontainer_m;
```

alpine/LoadBalancer.hpp
mohsensadr: I agree. @srikrrish @matt-frey do you agree with putting LoadBalancer in the src/Manger/ directory?

srikrrish: Yes if it is a generic enough code which can be used across applications it can be in src to avoid code repetition.

alpine/BumponTailInstabilityManager.h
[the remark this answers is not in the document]
mohsensadr: Something like this?
b0c0a86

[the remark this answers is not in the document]
srikrrish: May be `AlpineManager` is a better name than `PicManagerCommon`.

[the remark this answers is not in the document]
Arc676: I agree regarding the naming, since this is specific to ALPINE. But yes, this is what I meant.

alpine/ParticleContainer.hpp
Arc676: I think this is largely an issue of what we want the PIC manager to represent. As I mentioned in the Matrix channel, the PIC manager in IPPL core currently does nothing. For this reason, all of its members are also unnecessary restrictions (i.e. why does PIC manager care what data type the particle container is? It doesn't do anything with the particles). ALPINE is forced to propagate a data type into IPPL core in order to use that data type at application level.

Instead, the PIC manager could take the particle layout as its template parameter (as the original `ChargedParticles` did) and have a `std::vector<ParticleBase<PLayout>*>` member. This is more semantically meaningful, as a simulation does not necessarily have just one particle type, nor should the generic PIC manager depend on the exact particle properties. The vector contains pointers because the PIC manager base class does not own the particles. It merely acts on them (or will, eventually). Instead, the particles belong to the subclass (i.e. `class AlpineManager : public PICManager`), where the particle bunch has a data type that extends `ParticleBase` in order to describe a specific particle type.

In the following skeleton code, IPPL's core PIC manager is unaware of the particle properties. At the application level, the user defines their own particle type and extends the manager class to create a new manager class specific to a simulation containing bunches of that particle. The user's manager class is then free to access the specific particle properties as needed and the core PIC manager does not need access to the user's particle type.

```cpp
template <typename PLayout>
class PICManager {
  std::vector<ParticleBase<PLayout>*> particles;
};

class TachyonBunch : public ParticleBase<...> { ... };

template <...>
class FTLManager : public PICManager<ParticleSpatialLayout<...>> {
  TachyonBunch tachyons;

  FTLManager() : tachyons(...) { this->particles.push_back(&tachyons); }
};
```

src/Manager/PicManager.h
```diff
@@ -20,19 +20,37 @@
 
         virtual void grid2par() = 0;
 
-        void setParticleContainer(std::shared_ptr<pc> pcontainer){
+        virtual void dump() = 0;
```
Arc676, line 23: 1. Some documentation on the pure virtual methods would be useful.
2. Any specific reason the methods are called `grid2par` instead of `gather`? As far as I'm aware, the names of the PIC steps are well established.
3. Is the `dump` method necessary at IPPL core level? Having this as a pure virtual function forces the user to implement it, which means we are assuming things about the user's desired data output features.

alpine/BumponTailInstabilityManager.h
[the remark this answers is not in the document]
mohsensadr: done in 39d6f6d

src/Manager/PicManager.h
mohsensadr: 1. added fde22e6
2. This was suggested in the early template for pic manager by @matt-frey . I actually think grid2par/par2grid are general and self explanatory compared to gather/scatter which may be only known to the pic community.
3. I wanted to put post_step also in AlpineManager to avoid code repetition for mini apps. There, we dump the data at the end of each step. Although the implementation of dump is application specific, the idea of dumping data at the end of each step (or maybe every n steps) is universal among all applications I think and hence it makes sense. @srikrrish @matt-frey what do you think? should I remove it?

srikrrish: I think `grid2par` is ok as Mohsen said it is a bit generic than `gather` and hence we can just keep it. I am fine with either moving the `dump` from `PicManager` to `AlpineManager` or having a pure virtual function with a default implementation in `PicManager` itself. But I see that you have changed the pure virtual function to just a virtual function. Is that what @matt-frey you meant with default implementation? because then the class is no more abstract right?

alpine/LoadBalancer.hpp
matt-frey: Sri and I thought the load balance class to be a generic base class for all particle load balancing methods. It should therefore not be in Alpine. However, don't put this into src/Manager.

src/Manager/PicManager.h
matt-frey: To clarify the terminology. If a class has a *pure virtual function*, it is an *abstract class*, i.e. you cannot create an object of this type. You can only declare pure virtual functions, i.e. no function definition. The function definition is provided in derived classes. Hence, pure virtual functions cannot have a default implementation in the abstract class. If you want to provide a default implementation, the function must be `virtual` only.

alpine/ParticleContainer.hpp
matt-frey: I disagree, the PICManager should not take the particle layout as a template argument. I see the PICManager as a class that can perform PIC simulations, i.e. it has a stepper, particle container (or more than one), field solver elements and maybe a particle load balancer.

src/Manager/PicManager.h
srikrrish: Ok thanks for the clarification.

Arc676: I am also in favor of moving it to the ALPINE manager. The generic PIC manager should only contain symbols that are relevant at library level. Even if the concept of a regular data dump is universal, the exact logic is application-specific. For this purpose, the user is free to override `post_step` as @mohsensadr wrote. In more concrete terms: if the PIC manager never calls the `dump` function, it shouldn't be part of the class; if the symbol is only accessed from user code, it might as well also be defined in user code (also means the user is free to use any name or signature).

alpine/ParticleContainer.hpp
matt-frey: I otherwise agree with above example.

src/Manager/FieldSolverBase.h
mohsensadr: @matt-frey @srikrrish Do you agree on removing FieldSolver and its base class entirely? Should I implement its variables/methods in the AlpineManager?

alpine/LoadBalancer.hpp
mohsensadr: @matt-frey OK. Where should I put it then?

alpine/FieldContainer.hpp
mohsensadr: I removed `inline` everywhere. See dedf58a

alpine/BumponTailInstabilityManager.h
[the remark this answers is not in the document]
mohsensadr: It seems to reproduce the results of PenningTrap for P3M and OPEN as well (tested on cpu, openmp with 4 threads and 4 gpus compared to master's results)

alpine/LandauDampingManager.h
mohsensadr: Something like this 8fa698a ?

alpine/LandauDampingManager.h
```diff
@@ -39,72 +37,70 @@ struct CustomDistributionFunctions {
...
+    std::array<bool, Dim> decomp_m;
 public:
-    size_type getTotalP() const { return totalP; }
+    size_type getTotalP() const { return totalP_m; }
 
-    void setTotalP(size_type totalP_) { totalP = totalP_; }
+    void setTotalP(size_type totalP_) { totalP_m = totalP_; }
 
-    int getNt() const { return nt; }
+    int getNt() const { return nt_m; }
```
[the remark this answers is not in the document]
mohsensadr: It was done in previous commit [b0c0a86](https://github.com/IPPL-framework/ippl/pull/240/commits/b0c0a86cc47de834fc356a9de68e58a7229946e8)
Now we are calling the common class AlpineManger

alpine/LandauDampingManager.h
mohsensadr: Done in 94ec049

alpine/PenningTrapManager.h
mohsensadr: Done here d36b610

src/Manager/PicManager.h
mohsensadr: I moved `dump` to AlpineManger af38feb
Which elements of PicManager/BaseManager should be pure virtual, and which should be virtual? As far as I understand, both can be overwritten, but the user is forced to implement pure virtual ones. Now, everything is virtual. Is it important for BaseManager or PicManager class to be abstract (all member methods/functions to be pure virtual)?

In PicManager

```
	virtual void par2grid() { /* default does nothing */ };
	virtual void grid2par() { /* default does nothing */ };
```

and in BaseManager

```
        virtual void pre_run() { /* default does nothing */}

	virtual void pre_step() { /* default does nothing */}
        
	virtual void post_step() { /* default does nothing */}

	virtual void advance() { /* default does nothing */}
```

mohsensadr:
> I think we should move UniformPlasmaTest.cpp, ChargedParticles.hpp, LandauDampingMixedExec.cpp and LandauDampingMixedPrecision.cpp to a separate new directory as having them here makes it confusing.

Done here 1ca0a23
@srikrrish Let me know if you prefer a different name for the directory rather than alpine0

src/Manager/PicManager.h
Arc676: I would say that, as a general rule, your understanding of the concept also accurately describes the thought process that determines which should be which.

Methods that "do nothing by default" are optional behavior, i.e. the program is correct with or without any nontrivial definition for them. The user is free to ignore them if they are not relevant.

In contrast, a pure virtual function must be implemented, which means that it represents a concept that is required for the program to work but whose exact nature depends on the program. The implementation is application dependent but the concept is general to all applications.

For PIC, the steps in the PIC loop are not optional (otherwise it wouldn't be PIC). These are therefore pure virtual functions, as the PIC cycle always involves each step but the user must decide what happens in those steps. In particular, since the generic PIC manager does not know the particle types, the user must specify which attributes are relevant to the PIC loop[^1]. In contrast, pre-/post-step setup and cleanup are optional behavior; we don't assume that every user needs to have them. Thus, their default implementation is empty.

[^1]: Strictly speaking, there is a way around this by specifying a vector of attributes and associated fields for which the PIC loop is employed (introduces some complications regarding generic types), but I would say that this is beyond the scope of this PR.

alpine/LandauDampingManager.h
srikrrish: yes looks good to me..

srikrrish:
Regarding the naming for the directory containing the examples without PicManager, at present I also don't have a good solution. But alpine0 sounds confusing. Also this additional directory can still be within the alpine directory itself as they are still alpine related examples. Currently I can only think of naming the directory as `ExamplesWithoutPicManager` as that is the only common thing among them. May be we just do it like that for now and decide on a better name later.

mohsensadr:
> Regarding the naming for the directory containing the examples without PicManager, at present I also don't have a good solution. But alpine0 sounds confusing. Also this additional directory can still be within the alpine directory itself as they are still alpine related examples. Currently I can only think of naming the directory as `ExamplesWithoutPicManager` as that is the only common thing among them. May be we just do it like that for now and decide on a better name later.

Done here  822aae2 and 8a65884

src/Manager/PicManager.h
mohsensadr: In 000198c, I made par2grid, grid2par, and advance pure virtual methods. I guess these are the minimum elements of a PIC code. Although, this is very subjective...

srikrrish:
@mohsensadr Thanks for incorporating the comments and suggestions. I think some of the (possibly) larger issues in this PR can be addressed separately in other issues in the order of their priority and timing constraints. Hence I am approving it at the moment.

srikrrish (approved):

alpine/LoadBalancer.hpp
matt-frey: I thought we have a directory called `LoadBalancer` (or similar). If not, we may do this. We can also do this in further optimisations as @srikrrish already posted below.

mohsensadr: I don't see dir called "LoadBalancer" in src. We have [Decomposition](https://github.com/IPPL-framework/ippl/tree/pic-manager/src/Decomposition). I guess it should go there. I agree. Let's do it in another PR.

PR #244, opened 2023-12-06
Fix comm size check

s-mayani:
In Communicate.h isend(), there is a check for message count exceeding INT_MAX. However, currently this check is only checking whether the position where we are writing is bigger than INT_MAX, but not whether the message size being written is bigger than INT_MAX. This is because actually `getSize()` gives the buffer position and not its capacity, so this should be changed to `getBufferSize()`.


matt-frey:
This PR also adds a missing argument to the get function of the RMA interface.

aaadelmann (approved):

PR #246, opened 2023-12-11
Zigzag cic

manuel5975p:
Add a scatter function to ParticleAttribute that takes an additional position attribute to perform current deposition
Add ZigZag deposition scheme to `CIC.h(pp)`

- `ZigzagScatterToPoint` checks whether the deposition index is inbounds.
- `fractional_part` returns what would be the fractional part of `modf`
- `ZigzagScatterToField` takes two world-space vectors as opposed to `ScatterToField` which takes grid-space vectors


Arc676 (changes requested):
CIC stands for cloud-in-cell, which is an interpolation scheme. If zigzag is a new interpolation scheme, it should be stored in a new file in `src/Interpolation`; I think this will be cleaner and easier to manage. Code that is common to both interpolation schemes may warrant a new shared file as well.

Perhaps there was an internal discussion where I was not involved, but if not, I would suggest that the interface for this be discussed. We know that we want IPPL to support multiple interpolation schemes. I'm not sure that simple function overloading is the correct way to address this. Do all interpolation schemes necessarily require different sets of arguments? If not, the interpolation scheme should be explicitly chosen and visible at the call site such that it's clear from the caller code which interpolation scheme is being used.

src/Interpolation/CIC.h
```diff
@@ -119,6 +119,41 @@ namespace ippl {
             const std::index_sequence<GatherPoint...>&, const View& view,
             const Vector<T, View::rank>& wlo, const Vector<T, View::rank>& whi,
             const Vector<IndexType, View::rank>& args);
+
+        template <unsigned long ScatterPoint, unsigned long... Index, typename T, unsigned Dim,
+                  typename IndexType>
+        KOKKOS_INLINE_FUNCTION constexpr int ZigzagScatterToPoint(
+            const std::index_sequence<Index...>&,
+            const typename ippl::detail::ViewType<ippl::Vector<T, 3>, Dim>::view_type& view,
```
Arc676, line 127: If the `3` is correct, there should be some documentation to explain why it's 3 and not `Dim`

src/Interpolation/CIC.h
```diff
@@ -119,6 +119,41 @@ namespace ippl {
...
+         * to 'to' into a Kokkos View using specified grid spacing and scaling factor. It utilizes
+         * Kokkos functions for efficient parallelization.
+         */
+        template <unsigned long... ScatterPoint, typename T, unsigned Dim,
+                  typename IndexType = size_t>
+        KOKKOS_INLINE_FUNCTION void ZigzagScatterToField(
+            const std::index_sequence<ScatterPoint...>&,
+            const typename ippl::detail::ViewType<ippl::Vector<T, 3>, Dim>::view_type& view,
+            Vector<T, Dim> from, Vector<T, Dim> to,
+            const Vector<T, Dim> hr /*Grid Spacing*/, T scale, NDIndex<Dim> lDom, int nghost);
```
Arc676, line 155: `hr` is already explained to be the grid spacing in the docstring.

src/Interpolation/CIC.hpp
```diff
@@ -63,7 +63,135 @@ namespace ippl {
...
+                  typename IndexType>
+        KOKKOS_INLINE_FUNCTION constexpr int ZigzagScatterToPoint(
+            const std::index_sequence<Index...>&,
+            const typename ippl::detail::ViewType<ippl::Vector<T, 3>, Dim>::view_type& view,
+            const Vector<T, Dim>& wlo, const Vector<T, Dim>& whi,
+            const Vector<IndexType, Dim>& args, const Vector<T, Dim>& val, T scale, const Vector<T, Dim>& hr, const NDIndex<Dim> lDom, int nghost, const Vector<T, Dim>& source) {
+            (void)hr;
+            (void)lDom;
+            (void)nghost;
+            (void)source;
```
Arc676, line 76: Is this just to avoid warnings about unused arguments? Why are there unused arguments?

src/Interpolation/CIC.hpp
```diff
@@ -63,7 +63,135 @@ namespace ippl {
...
+            if(!isinbound){
+                return 0;
+            }
+            for(unsigned int d = 0;d < Dim;d++){
+                typename ippl::detail::ViewType<ippl::Vector<T, Dim>,
+                                            Dim>::view_type::value_type::value_type* destptr =
+                    &(view(interpolationIndex<ScatterPoint, Index>(args)...)[d]);
+                Kokkos::atomic_add(
+                    destptr,
+                    scale * val[d] * (interpolationWeight<ScatterPoint, Index>(wlo, whi) * ...));
```
Arc676, line 92: I'm not sure there is much gained in readability by having `destptr` be a separate variable instead of just putting the expression in the function call.

src/Interpolation/CIC.hpp
```diff
@@ -63,7 +63,135 @@ namespace ippl {
...
+            const Vector<T, Dim> hr, T scale, const NDIndex<Dim> lDom, int nghost) {
+            
+            // Use utility functions
+            
+            using Kokkos::max;
+            using Kokkos::min;
+            using Kokkos::floor;
+
+            Vector<T, Dim> from_in_grid_coordinates;
+            Vector<T, Dim> to_in_grid_coordinates;
```
Arc676, line 130: Use consistent casing for variable names.

src/Interpolation/CIC.hpp
```diff
@@ -63,7 +63,135 @@ namespace ippl {
...
+                whi[i] = fractional_part((to[i] + relay[i]) * T(0.5) / hr[i]);
+            }
 
+            auto __ =
+                (ZigzagScatterToPoint<ScatterPoint>(std::make_index_sequence<Dim>{}, view, wlo,
+                                                     whi, toi_local, jcto, scale, hr, lDom, nghost, source2)
+                 ^ ...);
+
+            (void)_;
+            (void)__; // [[maybe_unused]] causes issues on certain compilers
```
Arc676, line 192: Does the variable appear unused if you reassign the first placeholder instead of creating a new one?

Arc676 (changes requested):

srikrrish:
> CIC stands for cloud-in-cell, which is an interpolation scheme. If zigzag is a new interpolation scheme, it should be stored in a new file in `src/Interpolation`; I think this will be cleaner and easier to manage. Code that is common to both interpolation schemes may warrant a new shared file as well.
> 
> Perhaps there was an internal discussion where I was not involved, but if not, I would suggest that the interface for this be discussed. We know that we want IPPL to support multiple interpolation schemes. I'm not sure that simple function overloading is the correct way to address this. Do all interpolation schemes necessarily require different sets of arguments? If not, the interpolation scheme should be explicitly chosen and visible at the call site such that it's clear from the caller code which interpolation scheme is being used.

I totally agree with these comments. The different interpolation schemes should not be just function overloads of scatter. They should be in separate functions with the naming which explicitly says the intepolation scheme and also in different files. Previously we only had CIC interpolation/deposition and hence scatter/gather was synonymous with it. But now there is a new scheme their names should be changed as well i.e. scatterCIC, gatherCIC etc. and in future once the particle-in-Fourier branch gets merged we will have more new interpolation schemes such as scatterPIF, gatherPIF etc. So we should be careful in designing how we may want to do this properly.

src/Interpolation/CIC.h
manuel5975p: it's wrong, thanks for poynting out

src/Interpolation/CIC.h
manuel5975p: ok

src/Interpolation/CIC.hpp
manuel5975p: you're right, probably not

src/Interpolation/CIC.hpp
manuel5975p: to my surprise the warning disappears, I renamed to dummy_ since double underscore lead are reserved anyway

manuel5975p:
Okay I addressed all comments except the `ParticleAttrib::scatter` naming one, how should it be named? Because Zigzag is still CIC, but for lines instead of points.

src/Interpolation/Zigzag.h
```diff
@@ -0,0 +1,64 @@
...
+         * @param[in] view The Kokkos View to scatter data into.
+         * @param[in] from The starting point for scattering.
+         * @param[in] to The ending point for scattering.
+         * @param[in] hr Grid spacing for scattering.
+         * @param[in] scale The scaling factor for scattering.
+         *
+         * @details This function performs zigzag interpolation to scatter data points from 'from'
+         * to 'to' into a Kokkos View using specified grid spacing and scaling factor. It utilizes
+         * Kokkos functions for efficient parallelization.
+         */
```
matt-frey, line 24: These comments (including input parameters) does not reflect the code.

src/Interpolation/Zigzag.hpp
```diff
@@ -0,0 +1,121 @@
...
+         *
+         * @tparam ScatterPoint The scatter points to be used.
+         * @tparam T The data type of the field and view.
+         * @tparam Dim The dimensionality of the field and view.
+         * @tparam IndexType The index type used for indexing.
+         * @param view The view to scatter into.
+         * @param from The starting position of the scatter operation.
+         * @param to The ending position of the scatter operation.
+         * @param hr The grid spacing.
+         * @param scale The scaling factor to apply during the scatter operation. This is usually
```
matt-frey, line 41: What about "ldom" and "nghost"?

src/Interpolation/Zigzag.h
```diff
@@ -0,0 +1,64 @@
...
+         *
+         * @tparam ScatterPoint A list of unsigned long values representing the scatter points.
+         * @tparam T The data type of the values being scattered.
+         * @tparam Dim The dimension of the scatter operation.
+         * @tparam IndexType The data type used for indexing.
+         * @param[in] view The Kokkos View to scatter data into.
+         * @param[in] from The starting point for scattering.
+         * @param[in] to The ending point for scattering.
+         * @param[in] hr Grid spacing for scattering.
+         * @param[in] scale The scaling factor for scattering.
```
matt-frey, line 47: Missing "ldom" and "nghost".

src/Particle/ParticleAttrib.h
```diff
@@ -119,9 +119,29 @@ namespace ippl {
...
+        //     This performs zigzag deposition!
+        template <typename Field, typename P2, bool volumetricallyCorrect = false>
+        void scatter(Field& f, const ParticleAttrib<Vector<P2, Field::dim>, Properties...>& pp1,
+                     const ParticleAttrib<Vector<P2, Field::dim>, Properties...>& pp2,
+                     T dt_scale) const;
+
+        //     // scatter the data from this attribute onto the given Field, using
+        //     // the given Position attribute
+        template <typename Field, typename P2>
+        void scatterVolumetricallyCorrect(
```
matt-frey, line 136: What do you mean by "volumetrically correct"?

matt-frey: Found it a few lines below.

matt-frey (changes requested):
Can you maybe add more explanation to this zigzag interpolation? As zigzag also uses CIC underneath, I think it should not be treated like CIC. The "simple" scatter method -- we already have, which unfortunately currently has CIC hard-coded -- has a fundamental structure/kernel that is common to all "simple" scatter routines, irrespective of the interpolation order (i.e. constant, linear, quadratic etc.).

Arc676:
> Okay I addressed all comments except the `ParticleAttrib::scatter` naming one, how should it be named? Because Zigzag is still CIC, but for lines instead of points.

> The "simple" scatter method -- we already have, which unfortunately currently has CIC hard-coded -- has a fundamental structure/kernel that is common to all "simple" scatter routines

The fact that there is shared code between the scatter methods pushes, in my opinion, towards an interface of the form `void scatter(Attributes, Field, Positions, Algorithm, Arguments)`, where `Algorithm` is an enum type that identifies the different interpolation strategies and `Arguments` is some kind of container type for additional variables used for different interpolations. One possibility for the arguments is `std::variant<ArgTypes...>*` where `ArgTypes...` are different `struct`s with properties for different interpolations and passing `nullptr` indicates that the interpolation requires no additional parameters.

A shared function allows for the common code to be reused. We would just need to adapt the kernel to the desired interpolation strategy.

manuel5975p:
For just CIC and Zigzag, this makes limited sense to me. Zigzag takes different arguments, so passing an Algorithm option can be incompatible with the passed arguments (like calling `scatter(x1, x2, Algo::CIC)` or something would be ill-formed. 
At any rate, it would be nice to have @matt-frey and @srikrrish give their inputs on how the scatter function should look, especialli since Sri has mentioned another interpolation.

aaadelmann:
In IPPL 1.0 we had a construct like this: 

this->Q.scatter(this->rho_m, this->R, IntrplCIC_t()); 

and the gather operation was orthogonal. IntrplCIC_t() was just a typedef

I personally like this construct!

srikrrish:
As Andreas mentioned the particle positions, what quantity to scatter and a field to scatter on are the bare minimum needed for a scatter function. In the particle-in-Fourier scatter and gather I mentioned I also need to specify additionally a precomputed shape function in the Fourier space as computing it every time would be costly. It looks somewhat like this https://github.com/IPPL-framework/ippl/blob/131-implement-electrostatic-particle-in-fourier/src/Particle/ParticleAttrib.hpp#L652 (the MPI communicator at the end can be avoided with the new MPI interface from Matthias). So I would say it more or less still conforms to the generic scatter and gather interface. But the zigzag scatter indeed seems to be taking very different set of arguments.

manuel5975p:
So should I just rename attrib::scatter(x1, x2, field) to attrib::scatterZZ or something like that?

aaadelmann:
And why not: 

attrib::scatter(x1, x2, field, scatterMethod)

and scatterMethod is: IntrplCIC_t() or IntrplNGP_t() or IntrplZigzag_t() 

What do I miss here?

manuel5975p:
Because as it is now, CIC would only take one positional argument, whereas Zigzag takes two

aaadelmann:
yes I understand this but why can we not select at completive one of two variants?



On 20 Dec 2023, at 19:54, manuel5975p ***@***.***> wrote:



Because as it is now, CIC would only take one positional argument, whereas Zigzag takes two

src/Interpolation/Zigzag.hpp
```diff
@@ -0,0 +1,121 @@
...
+            if (!isinbound) {
+                return 0;
+            }
+            for (unsigned int d = 0; d < Dim; d++) {
+                Kokkos::atomic_add(
+                    &(view(interpolationIndex<ScatterPoint, Index>(args)...)[d]),
+                    scale * val[d] * (interpolationWeight<ScatterPoint, Index>(wlo, whi) * ...));
+            }
+
+            return 0;
```
Arc676, line 27: The loop can be nested instead of having the early return. The return value itself is a dummy value anyway so it makes more sense to only return once.

unit_tests/TestUtils.h
```diff
@@ -160,7 +159,7 @@ struct TestParams {
     }
 };
 
-bool TestParams::skipSerialTests = true;
+bool TestParams::skipSerialTests = false;
```
Arc676, line 162: This change doesn't belong in this PR and it also makes the `--run-serial` flag useless, since the tests will never be skipped. I believe it was decided to remove this entirely anyway, so instead of changing the default values, all of this should just be removed (possibly not in this PR).

srikrrish:
This was closed automatically by accident as dev-3.2.0 was deleted. Please reopen a new PR from your branch.

PR #247, opened 2023-12-15
Field sum

manuel5975p:
- Implement sum,prod,min,max for VFields with MPI
- unittests pass
- Cache MPI Datatypes and Ops in global maps
Currently, every Op & Type combo gets its own MPI_Op, not sure if that's optimal

### DIFF
--- a/src/Field/BareField.hpp
+++ b/src/Field/BareField.hpp
@@ -17,12 +17,46 @@
 namespace Kokkos {
     template <typename T, unsigned Dim>
     struct reduction_identity<ippl::Vector<T, Dim>> {
-        KOKKOS_FORCEINLINE_FUNCTION static ippl::Vector<T, Dim> sum() { return ippl::Vector<T, Dim>(0); }
-        KOKKOS_FORCEINLINE_FUNCTION static ippl::Vector<T, Dim> prod() { return ippl::Vector<T, Dim>(1); }
-        KOKKOS_FORCEINLINE_FUNCTION static ippl::Vector<T, Dim> min() { return ippl::Vector<T, Dim>( std::numeric_limits<T>::max()); }
-        KOKKOS_FORCEINLINE_FUNCTION static ippl::Vector<T, Dim> max() { return ippl::Vector<T, Dim>(-std::numeric_limits<T>::max()); }
+        KOKKOS_FORCEINLINE_FUNCTION static ippl::Vector<T, Dim> sum() {
+            return ippl::Vector<T, Dim>(0);
+        }
+        KOKKOS_FORCEINLINE_FUNCTION static ippl::Vector<T, Dim> prod() {
+            return ippl::Vector<T, Dim>(1);
+        }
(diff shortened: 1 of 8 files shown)


src/Field/BareField.hpp
```diff
@@ -149,8 +159,8 @@ namespace ippl {
     }
 
     DefineReduction(Sum, sum, valL += myVal, std::plus)
-    DefineReduction(Max, max, if (myVal > valL) valL = myVal, std::greater)
-    DefineReduction(Min, min, if (myVal < valL) valL = myVal, std::less)
+    DefineReduction(Max, max, using Kokkos::max;valL = max(valL, myVal), std::greater)
+    DefineReduction(Min, min, using Kokkos::min;valL = min(valL, myVal), std::less)
```
Arc676, line 163: Considering that this is a single-line function body, the `using` alias provides no benefit. It's actually longer than if you just put `Kokkos::max`, which I would recommend.

src/Types/Vector.h
```diff
@@ -45,7 +45,7 @@ namespace ippl {
         
         Vector(const std::array<T, Dim>& a);
 
-	Vector(const std::array<std::vector<T>, Dim>& a);
+	    Vector(const std::array<std::vector<T>, Dim>& a);
```
Arc676, line 48: What happened here? Recommend rerunning `clang-format`.

Arc676 (changes requested):
`get_mpi_op<Op>(op)` is redundant because the compiler can deduce `Op` from `op`. This also has the side effect of forcing an unused parameter in `Operations.h` even though the function only needs the data type.

Otherwise I defer to whatever was chosen in internal discussions, since I haven't been following this development. I would recommend a review by @matt-frey, who has more experience with MPI.

src/Field/BareField.hpp
manuel5975p: problem with using Kokkos::max directly is that inhibits ADL for types that don't have a max overload in the Kokkos namespace (e.g. quads, bigfloats etc. in their own namespace)

src/Types/Vector.h
manuel5975p: ah I see now. `clang-format` actually did that for me; however there's a tab in this line (not from me). I'll fix it with spaces.

manuel5975p:
> get_mpi_op<Op>(op) is redundant because the compiler can deduce Op from op
I agree, I just left it as-is

src/Field/BareField.hpp
manuel5975p: > Considering that this is a single-line function body, the using alias provides no benefit

that's wrong, the benefit isn't brevity but possibility of ADL for overloads

manuel5975p: Scratch that, it is important: the reason is that for vector types, the max overload isn't in the Kokkos namespace but in ippl. Using Kokkos would break it for ippl::Vector<T, D>

aaadelmann:
Sri and Matthias need to be added to the reviewers list

aaadelmann (approved):

Arc676 (approved):

PR #250, opened 2023-12-20
Add cuda-compatible ippl::Tuple with test

manuel5975p:
Add `__device__`-compliant `ippl::Tuple` type, inspired by std::tuple

Currently only in one file `Types/Tuple.h`, since most functions are trivial one-liners.

Some examples:
```cpp
{
    ippl::Tuple<int, std::unique_ptr<int>> tuple1(17, std::make_unique<int>(5));
    ippl::Tuple<int, std::unique_ptr<int>> tuple2{5, nullptr};
    //tuple2 = tuple1 is ill-formed
    tuple2 = std::move(tuple1); //Moves correctly
    
    ippl::Tuple<int, double> values1{5, 10.0};
    ippl::Tuple<int, double> values2(0.5f, 1.0f); //Both work, 0.5f is casted to 0 (type int).

    std::cout << ippl::get<0>(values1 + values2); // 5
    std::cout << ippl::get<1>(values1 - values2); // 9.0
}
```
Arithmetic operators are only enable for tuples in which all types satisfy [is_arithmetic](https://en.cppreference.com/w/cpp/types/is_arithmetic)

Tests work for Serial, OpenMP and Cuda, including a `parallel_for` and a `parallel_reduce` loop.

A usecase for this is Boundary Conditions for FDTD (and also general BCs) since the axis along which the wave vanishes should be a template parameter.


test/types/TestTuple.cpp
```diff
@@ -0,0 +1,41 @@
+#include "Types/Tuple.h"
+#include <Kokkos_Macros.hpp>
+#include <iostream>
+#include <memory>
+#include <Ippl.h>
+//#include <Kokkos_Core.hpp>
+#define alwaysAssert(X) do{if(!(X)){std::cerr<<"Assertion " << #X << " failed: " << __FILE__ << ": " << __LINE__ << '\n'; exit(-1);}}while(false)
```
Arc676, line 7: I would split this macro over several lines for readability.

However, this program also feels more like a unit test. Instead of building your own `assert`, you could just use the ones already in GTest.

src/Types/Tuple.h
```diff
@@ -0,0 +1,336 @@
...
+     */
+
+    /*!
+     * @struct TupleImpl
+     * @brief Implementation details for the Tuple class.
+     * @tparam i Current index in the Tuple.
+     * @tparam N Total number of elements in the Tuple.
+     * @tparam Ts Types of elements in the Tuple.
+     */
+    template <std::size_t i, std::size_t N, typename... Ts>
```
Arc676, line 26: What do you gain from `N` being its own parameter? Isn't `N == sizeof...(Ts)` invariant?

src/Types/Tuple.h
```diff
@@ -0,0 +1,336 @@
+//
+// Class Vector
+//   Vector class used for vector fields and particle attributes like the coordinate.
```
Arc676, line 3: Forgot to update.

Arc676 (changes requested):
Could use some additional empty lines to space out some things but otherwise seems good.

test/types/TestTuple.cpp
manuel5975p: Ah yes, that was a bit of a hack, also I just went off tests/vector

src/Types/Tuple.h
manuel5975p: Ts... changes, although I think it can be done better.

Arc676: This should be addressed in the doxygen comment, i.e. indicate that `Ts...` is the types of the remaining elements and not all the types in the tuple.

src/Types/Tuple.h
manuel5975p: Done

src/Types/Tuple.h
manuel5975p: Added a comment

src/Types/Tuple.h
```diff
@@ -0,0 +1,336 @@
...
+     * @brief Implementation details for the Tuple class.
+     * @tparam i Current index in the Tuple.
+     * @tparam N Total number of elements in the Tuple.
+     * @tparam Ts Types of elements in the Tuple.
+     */
+    template <std::size_t i, std::size_t N, typename... Ts>
+    struct TupleImpl;
+    /*!
+     * @struct TupleImpl<i, N, T, R, Ts...>
+     * @brief Partial specialization of TupleImpl for handling non-terminal elements.
```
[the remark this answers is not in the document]
manuel5975p: What this struct is supposed to handle is the distinction between

- We need at least one more element (which will be a member)
- This is the last element


A simplified version would be this
```cpp
template<typename T, typename R, typename... Ts>
struct Tuple{
    //Handle two or more elements (Ts... can have size 0)
};
template<typename T>
struct Tuple{
    //Handle one element
};
```
The above example does not compile however. We need to declare it such that the case distinction is done via [partial template specializations](https://en.cppreference.com/w/cpp/language/partial_specialization).
```cpp
template<typename... Ts>
struct Tuple;

template<typename T, typename R, typename... Ts>
struct Tuple<T, R, Ts...>{
    //Handle two or more elements (Ts... can have size 0)
    //One way to do this is like that:
    T this_value;
    Tuple<R, Ts...> the_rest;
};
template<typename T>
struct Tuple<T>{
    //Handle one element
    T this_value; //And no more remaining.
};
```

aaadelmann (approved):

PR #252, opened 2024-01-11
Add a Null Solver that has no action

s-mayani:
A NullSolver class inheriting from the Poisson base class is added. The NullSolver class is convenient because it allows to maintain the structure/software design of the program, requiring the user to only change the solver type to NullSolver when no solve action is wanted instead of having to code if statements. This functionality in needed in some programs such as OPAL when we seek to not have any solve action in a simulation.


aaadelmann (approved):

src/PoissonSolvers/NullSolver.h
```diff
@@ -0,0 +1,58 @@
...
+#define IPPL_NULL_SOLVER_H
+
+#include <Kokkos_MathematicalConstants.hpp>
+
+#include "Types/ViewTypes.h"
+
+#include "FFT/FFT.h"
+#include "FieldLayout/FieldLayout.h"
+#include "Index/NDIndex.h"
+#include "Poisson.h"
```
matt-frey, line 18: Some of these headers are not needed.

test/solver/TestNullSolver.cpp
```diff
@@ -0,0 +1,140 @@
...
+KOKKOS_INLINE_FUNCTION double gaussian(double x, double y, double z, double sigma = 0.05,
+                                       double mu = 0.5) {
+    double pi        = Kokkos::numbers::pi_v<double>;
+    double prefactor = (1 / Kokkos::sqrt(2 * 2 * 2 * pi * pi * pi)) * (1 / (sigma * sigma * sigma));
+    double r2        = (x - mu) * (x - mu) + (y - mu) * (y - mu) + (z - mu) * (z - mu);
+
+    return prefactor * exp(-r2 / (2 * sigma * sigma));
+}
+
+int main(int argc, char* argv[]) {
```
matt-frey, line 37: I do not really see the purpose of this test. It does not check anything.

matt-frey (changes requested):

src/PoissonSolvers/NullSolver.h
s-mayani: Removed.

test/solver/TestNullSolver.cpp
s-mayani: I agree, it is a bit useless. I didn't know what other sort of test one could do apart from checking that a 0 output is giving correctly even if the input is non-zero. Do you have any suggestions? 
Or I we remove this test?

matt-frey: I would remove this test or make it much simpler. It seems a bit much for testing nothing really. But then you can also just make a unit test.

s-mayani: There are no unit tests for solvers. Should I create a directory for solver unit tests just for this?

matt-frey: Ok. I thought we have unit tests for solvers [which would make sense]. Maybe remove this test for the moment and we add solver unit tests in a separate PR.

s-mayani: Done.

src/PoissonSolvers/NullSolver.h
```diff
@@ -0,0 +1,51 @@
...
+#ifndef IPPL_NULL_SOLVER_H
+#define IPPL_NULL_SOLVER_H
+
+#include "Poisson.h"
+
+namespace ippl {
+
+    template <typename FieldLHS, typename FieldRHS>
+    class NullSolver : public Poisson<FieldLHS, FieldRHS> {
+        constexpr static unsigned Dim = FieldLHS::dim;
```
matt-frey, line 17: This seems unused.

s-mayani: Fixed.

matt-frey (approved):
Thanks!

PR #253, opened 2024-01-12
Fix compilation issues + add FFT warmup function

s-mayani:
This PR addresses the following:

- Compilation issues in test/FFT/TestCos1.cpp and in FFTOpenPoissonSolver.hpp after previous merge, due to some conflicts with changes in Communicate.
- Add a warmup() function in the FFT class which performs a dummy FORWARD and BACKWARD transform. Calling this will always ensure that the warmup step is done for both directions, which when not done causes performance issues.
- Add a comment explaining the windowing of the Green's function in the Vico-Greengard solver.
- Remove all MPI Barriers from FFTOpenPoissonSolver.hpp, which were unnecessary and causing slowdown.

### DIFF
--- a/src/FFT/FFT.h
+++ b/src/FFT/FFT.h
@@ -177,6 +177,12 @@ namespace ippl {
         using typename Base::heffteBackend, typename Base::workspace_t, typename Base::Layout_t;
 
         /*!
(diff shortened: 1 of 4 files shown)


matt-frey (approved):
Looks good to me! Thanks!

srikrrish (approved):

PR #256, opened 2024-01-31
Variant patch

### LINKED ISSUE #249: appending instead over overwriting CXX_Flags
Currently and in proposed changes, we overwrite existing compiler flags, see:

https://github.com/IPPL-framework/ippl/pull/242
https://github.com/IPPL-framework/ippl/blob/b19e33e34b9a1dab2f4ba051990b6fb01d2b5630/CMakeLists.txt#L40-L43

I do think we could save the duplication of the code, and also make it a bit more clear instead we only append the new flags to the existing ones over something like for example for CUDA debug flags:
```cmake
set (CMAKE_CXX_FLAGS_DEBUG  "${CMAKE_CXX_FLAGS_DEBUG} -G")
```
Also one benefit could be if one only wants to add more flags, however is in a CUDA build it should not matter where the flags are added, thus would be a bit save to properly have them set. 

Further information, or a small discussion from that also on stack overflow.

https://stackoverflow.com/questions/48462133/appending-cxx-flags-in-cmake-invocation

Arc676:
Adds an alternative implementation of `std::variant` that works under CUDA 12.2 with GCC 12.3.0. The alternative implementation is activating by configuring `-DUSE_ALTERNATIVE_VARIANT=ON` during CMake setup.

Closes #249, because there's no reason not to.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,7 +39,7 @@ set (CMAKE_CXX_FLAGS_RELWITHDEBINFO "-O3 -g ")
 set (CMAKE_CXX_FLAGS_RELEASE "-O3")
 set (CMAKE_CXX_FLAGS_DEBUG  "-O0 -g")
 if ("${Kokkos_DEVICES}" MATCHES "CUDA")
-    set (CMAKE_CXX_FLAGS_DEBUG "-O0 -g -G")
+    set (CMAKE_CXX_FLAGS_DEBUG  "${CMAKE_CXX_FLAGS_DEBUG} -G")
 endif()
 
 # Resolve all library dependencies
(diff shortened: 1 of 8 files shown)


srikrrish:
It seems `#pragma GCC system_header` is missing in the Variant

Arc676:
> It seems `#pragma GCC system_header` is missing in the Variant

https://github.com/IPPL-framework/ippl/pull/256/files#diff-09a83b51ac8b436d4ddd9263eb444c2fcd63bc8bddbd18cea24daed7646dc49eR33

srikrrish:
Ah ok.. i see it now.

srikrrish (approved):

s-mayani (approved):

aaadelmann (approved):

PR #258, opened 2024-02-26
Merge FEM framework for Poisson eq.

s-mayani:
This PR contains the building blocks for the Finite Element Method. Currently, it works only with Lagrange elements. These are sufficient to solve the Poisson equation, which is why there is a FEMPoissonSolver based on the FEM building blocks. There are unit tests and a test for the FEMPoissonSolver. 

Further work will work on this structure to add different types of elements (e.g. Nédélec).

Please note that currently this only works serially.


srikrrish:
This was closed automatically by accident as dev-3.2.0 was deleted. Hence reopening.

srikrrish:
This was closed automatically by accident as dev-3.2.0 was deleted. Hence reopening.

PR #261, opened 2024-03-13
Fix compilation issues & cleanup

manuel5975p:
- Remove Userlist (not requred, causes compilation error due to removed std::binary_function)
- Fix ctor inheritance for apple clang (c++20 aggregate initialization)
- workaround for static_assert(false) and add some comments.


aaadelmann:
Please describe above why we remove the UserList

aaadelmann (approved):

s-mayani (approved):

PR #262, opened 2024-03-15
Remove aggr. init. for Apple Clang

manuel5975p:
Explicitly add constructors to TupleImpl<i, N, T>
```cpp
        TupleImpl<i, N, T>()
            requires(std::is_default_constructible_v<T>)
         = default;
        TupleImpl<i, N, T>(const T& t) : val(t){}
        TupleImpl<i, N, T>(T&& t) : val(std::forward<T>(t)){}
```
in order not to rely on C++20's aggregate initialization since apple clang [doesn't implement that feature](https://en.cppreference.com/w/cpp/compiler_support/20)
The red cell is Apple Clang's support for this feature.
image

### DIFF
--- a/src/Types/Tuple.h
+++ b/src/Types/Tuple.h
@@ -105,6 +105,11 @@ namespace ippl {
         KOKKOS_INLINE_FUNCTION const auto& get() const noexcept {
             return val;
         }
+        TupleImpl<i, N, T>()
+            requires(std::is_default_constructible_v<T>)
+         = default;
+        TupleImpl<i, N, T>(const T& t) : val(t){}
+        TupleImpl<i, N, T>(T&& t) : val(std::forward<T>(t)){}
     };
     /*!
      * @class Tuple


Arc676:
The changes seem fine but the PR is lacking in motivation. What is the benefit of adding this constructor? The custom tuple implementation is only relevant for machines running GCC12 + CUDA 12.1, which implies not using Apple Clang. Does IPPL fail to compile on Macs when using the standard library tuple implementation?

manuel5975p:
I just don't want apple clang users to get bogged down by the TestTuple executable, which doesn't make sense / work with std::tuple

Arc676 (commented):
Never mind, got confused. Was thinking of the variant type.

Arc676 (approved):

Arc676 (changes requested):
Actually, on second thought, if this is a compiler-specific fix that deviates from the STL implementation (I assume most of this code is adapted from GNU stdlibc++) just because of implementation choices in Clang, I would surround this with an `#ifdef` to enable this change only for the affected compiler.

Arc676 (approved):

aaadelmann (approved):

PR #264, opened 2024-03-18
Add step method to alpine headers

s-mayani:
In the Alpine tests there is the step_method missing from the argument list in the headers.

### DIFF
--- a/alpine/BumponTailInstability.cpp
+++ b/alpine/BumponTailInstability.cpp
@@ -1,8 +1,8 @@
 // Bump on Tail Instability Test Case
 //   Usage:
 //     srun ./BumponTailInstability
-//                  <nx> [<ny>...] <Np> <Nt> <stype>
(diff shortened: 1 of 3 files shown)


aaadelmann (approved):

srikrrish (changes requested):
What is the use of this? We only have Leap frog or Boris at the moment and it seems to unnecessarily increase the already long list of command line arguments

s-mayani:
> What is the use of this? We only have Leap frog or Boris at the moment and it seems to unnecessarily increase the already long list of command line arguments

I do not know what the use of this is; it was added by the Pic Manager merge. I just realised that it was inconsistent in the headers (not explained) and Felix and Anna's documentation also propagated this error.

srikrrish:
Oh I see.. Sorry didn't realize you just changed only the comments. I think this question should have been asked in the Pic Manager merge itself. Anyway since it is there already I will just approve it..

srikrrish (approved):

PR #268, opened 2024-03-20
Change MPI tags

s-mayani:
Change MPI tags in IPPL, as they are causing issues on big node runs (issue observed in FFTOpenPoissonSolver.hpp). 
Also removed unnecessary tags.

### DIFF
--- a/src/Communicate/Tags.h
+++ b/src/Communicate/Tags.h
@@ -9,87 +9,36 @@ namespace ippl {
     namespace mpi {
         namespace tag {
             enum Tag : int {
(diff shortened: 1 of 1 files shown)


aaadelmann (approved):
I commit but I think we need to rethink this. I have a feeling it is only a matter of time if we run into the very same problem.

matt-frey:
@aaadelmann Sonali and I are aware of this. We just wanted to make a temporary fix (for the coming release) such that we can use this for the Vico paper.

srikrrish (approved):

PR #281, opened 2024-04-19
Rewrite CMakeLists.txt

manuel5975p:
## Synopsis

- Packages `heffte` and `kokkos` are looked for locally, if not found `FetchContent` is used to download
- `IPPL_PLATFORMS` can be set to one of the following: 
  - `serial`
  - `openmp`
  - `cuda`
  - `openmp;cuda` or `cuda;openmp`
  - empty defaults to `serial`
- Platform `cuda` implies the option `Heffte_ENABLE_CUDA`
- Other platforms imply `Heffte_ENABLE_AVX2`, if use of FFTW is desired set `Heffte_ENABLE_FFTW`

---

## Explanation of different sections of code:
```cmake
if(NOT WIN32)
  string(ASCII 27 Esc)
  set(ColourReset "${Esc}[m")
  set(Red         "${Esc}[31m")
  set(Green       "${Esc}[32m")
  set(Yellow      "${Esc}[1;33m")
endif()
```

Sets a couple of color codes to highlight important output parts for warnings and hints.

---
```cmake
if(NOT IPPL_PLATFORMS)
    message(STATUS "${Yellow}IPPL_PLATFORMS flag is not set, defaulting to serial backend${ColourReset}")
    set(IPPL_PLATFORMS "SERIAL")
endif()
if(IPPL_PLATFORMS)
    string(TOUPPER "${IPPL_PLATFORMS}" IPPL_PLATFORMS)
    #Handle
endif()
```
Defaults `IPPL_PLATFORMS` to SERIAL, also converts `IPPL_PLATFORMS` to uppercase to make options case insensitive.


srikrrish:
Looks good in general. But I have some questions. How do the architecture for Kokkos or other Kokkos, Heffte options are set in this? Also it seems to always include Heffte without considering the ENABLE_FFT flag is that right?

felixschurk:
I guess I should not make a review :D

manuel5975p:
Heffte is neither looked for nor fetched if ENABLE_FFT is not set, the `add_definitions` line is not from me. 
```cmake
if (ENABLE_FFT)
    add_definitions (-DENABLE_FFT)
    find_package (Heffte 2.2.0 REQUIRED)
    find_package (Heffte 2.2.0 QUIET)
    if(NOT Heffte_FOUND)
        message(STATUS "Using FetchContent for Heffte")
        FetchContent_Declare(
             Heffte
             #DOWNLOAD_EXTRACT_TIMESTAMP
             GIT_REPOSITORY https://github.com/icl-utk-edu/heffte 
        )
        FetchContent_MakeAvailable(Heffte)
    endif()
    message (STATUS "Found Heffte_DIR: ${Heffte_DIR}")
endif ()
```

---

The Kokkos and Heffte architectures are set as follows:
```cmake
    if("${IPPL_PLATFORMS}" STREQUAL "CUDA")
        message(STATUS "${Green}CUDA backend enabled${ColourReset}")
        set(Kokkos_ENABLE_CUDA ON CACHE BOOL "Enable Kokkos CUDA Backend")
        set(Heffte_ENABLE_CUDA ON CACHE BOOL "Set Heffte CUDA Backend")

# And so on...
```
Kokkos and Heffte are both configurable via cmake options if they are part of the project.

CMakeLists.txt
```diff
@@ -1,47 +1,109 @@
...
+endif()
 if (${CMAKE_BUILD_TYPE} STREQUAL "Release")
     add_definitions (-DNOPAssert)
 endif ()
 
 add_compile_options (-Wall)
 add_compile_options (-Wunused)
 add_compile_options (-Wextra)
-add_compile_options (-Werror)
+#add_compile_options (-Werror)
```
srikrrish, line 38: Why is this commented?

test/types/CMakeLists.txt
```diff
@@ -19,12 +19,12 @@ target_link_libraries (
     ${IPPL_LIBS}
     ${MPI_CXX_LIBRARIES}
 )
-add_executable (TestTuple TestTuple.cpp)
-target_link_libraries (
-    TestTuple
-    ${IPPL_LIBS}
-    ${MPI_CXX_LIBRARIES}
-)
+#add_executable (TestTuple TestTuple.cpp)
```
srikrrish, line 22: Same here why is this commented?

srikrrish (changes requested):

srikrrish:
> Heffte is neither looked for nor fetched if ENABLE_FFT is not set, the `add_definitions` line is not from me.
> 
> ```cmake
> if (ENABLE_FFT)
>     add_definitions (-DENABLE_FFT)
>     find_package (Heffte 2.2.0 REQUIRED)
>     find_package (Heffte 2.2.0 QUIET)
>     if(NOT Heffte_FOUND)
>         message(STATUS "Using FetchContent for Heffte")
>         FetchContent_Declare(
>              Heffte
>              #DOWNLOAD_EXTRACT_TIMESTAMP
>              GIT_REPOSITORY https://github.com/icl-utk-edu/heffte 
>         )
>         FetchContent_MakeAvailable(Heffte)
>     endif()
>     message (STATUS "Found Heffte_DIR: ${Heffte_DIR}")
> endif ()
> ```
> 
> The Kokkos and Heffte architectures are set as follows:
> 
> ```cmake
>     if("${IPPL_PLATFORMS}" STREQUAL "CUDA")
>         message(STATUS "${Green}CUDA backend enabled${ColourReset}")
>         set(Kokkos_ENABLE_CUDA ON CACHE BOOL "Enable Kokkos CUDA Backend")
>         set(Heffte_ENABLE_CUDA ON CACHE BOOL "Set Heffte CUDA Backend")
> 
> # And so on...
> ```
> 
> Kokkos and Heffte are both configurable via cmake options if they are part of the project.

Ok thanks

CMakeLists.txt
```diff
@@ -1,47 +1,109 @@
...
+        # Handle unknown or unsupported platform
+    endif()
+endif()
 
+find_package(Kokkos 4.0.0 QUIET)
+if(NOT Kokkos_FOUND)
+    FetchContent_Declare(
+        #DOWNLOAD_EXTRACT_TIMESTAMP
+        kokko
+        URL https://github.com/kokkos/kokkos/archive/refs/tags/4.1.00.tar.gz
```
matt-frey, line 91: I see that this is hard-coded to fetch Kokkos version 4.1.00. Can't we make this more general?

CMakeLists.txt
```diff
@@ -73,7 +135,16 @@ endif()
...
 if (ENABLE_FFT)
     add_definitions (-DENABLE_FFT)
-    find_package (Heffte 2.2.0 REQUIRED)
+    find_package (Heffte 2.2.0 QUIET)
+    if(NOT Heffte_FOUND)
+        message(STATUS "Using FetchContent for Heffte")
+        FetchContent_Declare(
+             Heffte
+             #DOWNLOAD_EXTRACT_TIMESTAMP
+             GIT_REPOSITORY https://github.com/icl-utk-edu/heffte
```
matt-frey, line 144: What version of HeFFTe will this fetch? I think we should be able to specify the version.

test/types/CMakeLists.txt
matt-frey: Didn't we comment this out because the tuple is not yet properly working?

matt-frey (changes requested):
I like the idea, but I still have a few questions. In addition to my comments, I would like to know
* How does one specify other Kokkos or HeFFTe flags when configuring Ippl?
* What happens when a Kokkos build or HeFFTe build fails? Will it abort the whole configure process?

CMakeLists.txt
manuel5975p: This will fetch master, which is required for `TestCos1` afaik, right @vmontanaro?

CMakeLists.txt
manuel5975p: We can do that via variable:
```
set(KOKKOS_VERSION "4.1.00")

# Use the variable in FetchContent_Declare
FetchContent_Declare(
    kokkos
    URL https://github.com/kokkos/kokkos/archive/refs/tags/${KOKKOS_VERSION}.tar.gz
)
```
or more sophisticated via major and minor? Let me know your thoughts.

matt-frey: I think passing the Kokkos version like this is fine. However, we must check if the file exists.

CMakeLists.txt
vmontanaro: Correct.

CMakeLists.txt
```diff
@@ -1,47 +1,109 @@
...
+        message(FATAL_ERROR "Unknown or unsupported platform: ${IPPL_PLATFORMS}")
+        # Handle unknown or unsupported platform
+    endif()
+endif()
 
+find_package(Kokkos 4.0.0 QUIET)
+if(NOT Kokkos_FOUND)
+    FetchContent_Declare(
+        #DOWNLOAD_EXTRACT_TIMESTAMP
+        kokko
```
Arc676, line 90: Typo.

Arc676 (approved):

CMakeLists.txt
manuel5975p: I actually named it that on purpose to avoid name duplication (fetchtcontent object vs. target name), but I think that's superfluous.

CMakeLists.txt
manuel5975p: So I guess since there's no release with Cos1, we're gonna have to stick with master.

srikrrish: Yes for now that's fine.. But just like passing Kokkos version as an option here also passing Heffte version as an option would be good. It can default to master if none is passed.

CMakeLists.txt
manuel5975p: I addressed it: specifying no kokkos version yields:
image
Otherwise: `cmake .. -DKOKKOS_VERSION=4.2.00` yields:
image

---

For heffte:
Default behaviour (`cmake .. -DENABLE_FFT=True`) is again:
image
For a specified version `cmake .. -DENABLE_FFT=True -DHEFFTE_VERSION=2.4.0`:
image

manuel5975p:
Addressed the commetns

CMakeLists.txt
matt-frey: Yes, that is fine. I am just wondering about wrong inputs. What happens if a release does not exist? What does `FetchContent_Declare` do in such a case?

manuel5975p: Currently it just prints a verbose "download failed" message

srikrrish:
Please modify the readme explaining how to do a serial, openmp and cuda builds. I can approve then.

CMakeLists.txt
matt-frey: Does cmake print this message and abort?

manuel5975p: Yes. To be precise, `FetchContent_MakeAvailable` attempts a https download and the github server responds with an error message `22`, for example:
```
    error: downloading 'https://github.com/kokkos/kokkos/archive/refs/tags/4.43.00.tar.gz' failed
          status_code: 22
          status_string: "HTTP response code said error"
          log:
          --- LOG BEGIN ---
          Host github.com:443 was resolved.
```

matt-frey: OK. When cmake aborts, this is fine for me.

manuel5975p:
Readme is updated

README.md
```diff
@@ -6,22 +6,70 @@ Independent Parallel Particle Layer (IPPL) is a performance portable C++ library
...
+```
+cmake <src_dir> -D<option>=<value>
+```
+#### None of the options have to be set explicitly, all have a default.
+
+The relevant options of IPPL are
+- IPPL_PLATFORMS, can be one of `SERIAL, OPENMP, CUDA, "OPENMP;CUDA"`, default `SERIAL`
+- `KOKKOS_VERSION`, default `4.1.00`
+- `HEFFTE_VERSION`, default `MASTER`
+- `HEFFTE_VERSION`, can be default `MASTER`
```
srikrrish, line 35: Is this appearing twice?

README.md
```diff
@@ -6,22 +6,70 @@ Independent Parallel Particle Layer (IPPL) is a performance portable C++ library
...
+- `ENABLE_SOLVERS`, default `OFF`
+- `ENABLE_FFT`, default `OFF`
+  - If `ENABLE_FFT` is set, `Heffte_ENABLE_CUDA` will default to `ON` if `IPPL_PLATFORMS` contains `cuda`
+  - Otherwise, `Heffte_ENABLE_AVX2` is enabled. FFTW has to be enable explicitly.
+- `Heffte_ENABLE_FFTW`, default `OFF`
```
srikrrish, line 40: Does it download and install FFTW if the option is ON?

srikrrish (changes requested):

README.md
manuel5975p: It doesnt download fftw, FFTW is not cmake based afaik, I can arrange some setup if you want but it's hacly

README.md
manuel5975p: oops

README.md
srikrrish: That's ok.. I don't think we need to do that. The user can/should install FFTW separately. So I guess if the flag is on but heffte can't find FFTW then heffte itself will print the error message right? (like FFTW_DIR not found or something like that). If that is the case we don't need to do anything and leave it as it is..

manuel5975p: Yes, the configuration of heffte fails if fftw isnt available

manuel5975p:
Duplication is removed

CMakeLists.txt
matt-frey: Maybe add a comment and then resolve this.

README.md
```diff
@@ -6,22 +6,69 @@ Independent Parallel Particle Layer (IPPL) is a performance portable C++ library
...
+mkdir build
+cd build
+```
+#### Serial debug build with tests and newest Kokkos
+```
+cmake .. -DCMAKE_BUILD_TYPE=Debug -DENABLE_TESTS=True -DKOKKOS_VERSION=4.2.00
+```
+#### OpenMP release build with tests and FFTW
+```
+cmake .. -DCMAKE_BUILD_TYPE=Release -DIPPL_PLATFORMS=openmp -DHeffte_ENABLE_FFTW -DENABLE_TESTS=True
```
srikrrish, line 63: Don't you have to point to FFTW installation directory through heffte options in this case?

srikrrish (changes requested):

README.md
manuel5975p: I think my cmake and the 200-build-heffte script both rely on Heffte to find fftw, and it works for me locally and on merlin.

srikrrish: Yeah but in the build scripts we do specify this https://github.com/IPPL-framework/ippl-build-scripts/blob/main/200-build-heffte#L60C5-L60C30 . Are you also specifying it somewhere?

srikrrish: You mean `FFTW_DIR`? If you install FFTW as a module commonly that environment variable is also set. Otherwise you have to manually set it before executing the script.

manuel5975p: Ah yes, then it's picked up by cmake as well.

srikrrish: So are you saying we don't need to explicitly specify this `-DFFTW_ROOT="${FFTW_DIR}"` cmake option in this line and cmake automatically picks it up?

manuel5975p: I just corrected this example from `-DHeffte_ENABLE_FFTW` to `-DHeffte_ENABLE_FFTW=True`, and all I can say with certainty that `cmake .. -DCMAKE_BUILD_TYPE=Release -DIPPL_PLATFORMS=openmp -DHeffte_ENABLE_FFTW=True -DENABLE_TESTS=True -DENABLE_FFT=True -DENABLE_SOLVERS=True` and ` TestFFTPeriodicPoissonSolver` compiles.

manuel5975p:
So, whats the current concern?

srikrrish:
> So, whats the current concern?

Nothing from my side but I still see a few more unresolved threads. As Andreas mentioned it is also good to test this in a few different systems and confirm it is working. I can test it in Juwels CPU and GPUs until the end of today. @s-mayani Is it possible for you to test this on Perlmutter?

s-mayani:
> > So, whats the current concern?
> 
> Nothing from my side but I still see a few more unresolved threads. As Andreas mentioned it is also good to test this in a few different systems and confirm it is working. I can test it in Juwels CPU and GPUs until the end of today. @s-mayani Is it possible for you to test this on Perlmutter?

Yes, I can test this on Perlmutter.

srikrrish:
Also can we check out a specific SHA of Heffte with the cmake options. Because for the DCT reasons we have to use the Heffte master currently and that is not good in terms of reproducibility.

manuel5975p:
I'll add the git hash later today, I know fetchcontent supports it.

srikrrish:
> I'll add the git hash later today, I know fetchcontent supports it.

That would be great! Thanks

test/types/CMakeLists.txt
manuel5975p: Yes

manuel5975p:
Okay, I tried the formerly error-generating configurations and it seems to work locally and on merlin. I also changed some lines in [test/random/TestInverseTransformSamplingCustom.cpp](https://github.com/IPPL-framework/ippl/pull/281/commits/34ba7aedc169d1dddf902d00bd6e3228842ab242#diff-9912dc49e96ed73a5ad8f87d8a94664cdf65755e409bcfe6c21e2390e2b8852b) since it generated a warning (due to being technically incorrect C++ code). It should also work for clang-based builds.

CMakeLists.txt
manuel5975p: changed it

srikrrish (approved):

manuel5975p:
> I like the idea, but I still have a few questions. In addition to my comments, I would like to know
> 
>     * How does one specify other Kokkos or HeFFTe flags when configuring Ippl?
> 
>     * What happens when a Kokkos build or HeFFTe build fails? Will it abort the whole configure process?

To answer these question:
1. By simply passing in the option that their CMakeLists.txt tests for
2. Somewhat yes, however in cases where a target doesn't depend on e.g. Heffte, it's buildable even if Heffte fails. Heffte and Kokkos objects are just inserted into the make tree as if it was one project.

manuel5975p:
@matt-frey it still says "matt-frey requested changes", are those changes addressed?

matt-frey:
If a dependency does not build upon CMake, we cannot use this approach, right?

matt-frey:
Does it work? I know that @srikrrish tried to compile and had several issues, are these resolved?

srikrrish:
> Does it work? I know that @srikrrish tried to compile and had several issues, are these resolved?

Yes it now works on Juwels GPU, OMP and serial builds. There still seem to be some warnings from serial build but even with build scripts they occur and unrelated to this cmake change. So there will be a new issue opened with regards to that.

manuel5975p:
> If a dependency does not build upon CMake, we cannot use this approach, right?

There are ways to handle non cmake-based, and at any rate find_package works regardless. In the absolute, absolute worst case, we're back at running a script from cmake again.

matt-frey (approved):

PR #287, opened 2024-05-01
Changing Template variable in PicManager

JoMee:
I changed the template variable of the field solver class to be more generic. One needs this to for example run multiple solvers in one step. I don't see a reason why this is not generic, I also updated the alpine manager and the simulations ran without problem.

### DIFF
--- a/alpine/AlpineManager.h
+++ b/alpine/AlpineManager.h
@@ -8,6 +8,7 @@
 #include "LoadBalancer.hpp"
 #include "Manager/BaseManager.h"
 #include "Manager/PicManager.h"
+#include "Manager/FieldSolverBase.h"
 #include "ParticleContainer.hpp"
 #include "Random/Distribution.h"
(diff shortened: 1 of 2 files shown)


manuel5975p:
I could be wrong, but if the pointer type is going to be a template parameter why would we not remove the polymorphism altogether? Going from
`std::shared_ptr<solver_base>` to simply a `solver_type`?

JoMee:
That depends if we want to support switching out the solver at runtime in the future (as we currently do with the FieldContainer or ParticleContainer). I don't think the overhead of this attribute is going to be too large. 
Another point is if we want to return a pointer to the field solver at some point then it would be correct to have it as a shared pointer. (Since if the PICManager is destructed, the field solver would also get destructed)

In summary, it is probably the more future-proof way and doesn't add significant overhead, this is why I would keep it in a shared_ptr.

PR #290, opened 2024-05-06
Change Maxwell class to take 4-current

s-mayani:
Changed the Maxwell base class to take four current (charge density and current density in same vector) instead of two separate charge and current density fields.

### DIFF
--- a/src/MaxwellSolvers/Maxwell.h
+++ b/src/MaxwellSolvers/Maxwell.h
@@ -18,21 +18,20 @@ namespace ippl {
     template <typename EMField, typename SourceField>
     class Maxwell {
     public:
-        // typedefs for the different fields and vector fields
-        using typeR = typename SourceField::value_type;
-        using typeL = typename EMField::value_type;
-
         constexpr static unsigned Dim = EMField::dim;
-
-        // type for communication buffers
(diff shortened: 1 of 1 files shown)


src/MaxwellSolvers/Maxwell.h
```diff
@@ -18,46 +18,35 @@ namespace ippl {
     template <typename EMField, typename SourceField>
     class Maxwell {
```
manuel5975p, line 19: No inheritance from SolverBase?
```mermaid
graph TB

X[SolverBase] --> B(Electrostatics)
X[SolverBase] --> A(Maxwell)
B[Electrostatics] --> C[Poisson]
B[Electrostatics] --> FFT
B[Electrostatics] --> FEM
A(Maxwell) --> D(FDTD)
A(Maxwell) --> E(FEM)
```

manuel5975p (approved):

src/MaxwellSolvers/Maxwell.h
aaadelmann: I feel SolverBase would be over design. However, we can discuss this later.

aaadelmann (approved):

PR #296, opened 2024-05-23
Update README.md to add jobscripts for Merlin

s-mayani:


aaadelmann (approved):

matt-frey (approved):
This looks okay. However, I would personally not put such things into the ippl repository. We could have a separate repo for this which is private to us. External users are not interested in PSI local clusters.

s-mayani:
> This looks okay. However, I would personally not put such things into the ippl repository. We could have a separate repo for this which is private to us. External users are not interested in PSI local clusters.

For now, it seems like since most people using IPPL are running on the PSI clusters, this makes sense to me.

aaadelmann:
I agree with Sonali. While this is maybe not the cleanest way, we are constantly struggling with such issues just because there is no place to look for. On the other hand I do not want to have an inflation of repos, we do not have enough manpower. 
I would argues tat  external user can easy adapt to their cluster hence there is an interest.

PR #307, opened 2024-08-21
Updating Inverse Transform Sampling class

mohsensadr:
Allow updating bounds and local number of particles.

### DIFF
--- a/src/Random/InverseTransformSampling.h
+++ b/src/Random/InverseTransformSampling.h
@@ -37,28 +37,38 @@ namespace ippl {
         /*!
          * @brief Constructor for InverseTransformSampling class with domain decomposition.
          *
-         * @param dist_ The distribution to sample from.
-         * @param rmax_ Maximum range for sampling.
-         * @param rmin_ Minimum range for sampling.
-         * @param rlayout The region layout.
-         * @param ntotal_ Total number of samples to generate.
+         * @param dist_r The distribution to sample from.
+         * @param rmax_r Maximum global range.
+         * @param rmin_r Minimum global range.
+         * @param rlayout_r The region layout.
+         * @param ntotal_r Total number of samples to generate.
         */
         template <class RegionLayout>
         InverseTransformSampling(Distribution &dist_r, Vector<T, Dim> &rmax_r, Vector<T, Dim> &rmin_r, RegionLayout& rlayout_r, size_type &ntotal_r)
         : dist_m(dist_r),
         ntotal_m(ntotal_r){
 
-            updateBounds(rmax_r, rmin_r, rlayout_r);
+            const typename RegionLayout::host_mirror_type regions = rlayout_r.gethLocalRegions();
+            int rank = ippl::Comm->rank();
+
+            Vector<T, Dim> locrmax, locrmin;
+            for (unsigned d = 0; d < Dim; ++d) {
+                locrmax[d] = regions(rank)[d].max();
+                locrmin[d] = regions(rank)[d].min();
+            }
+
+            updateBounds(rmax_r, rmin_r, locrmax, locrmin);
 
         }
 
         /*!
          * @brief Constructor for InverseTransformSampling class without applying domain decomposition..
          *
-         * @param dist_ The distribution to sample from.
-         * @param rmax_ Maximum range for sampling.
-         * @param rmin_ Minimum range for sampling.
-         * @param rlayout The region layout.
+         * @param dist_r The distribution to sample from.
+         * @param rmax_r Maximum global range.
+         * @param rmin_r Minimum global range.
+         * @param locrmax_r Maximum local (per rank) range.
+         * @param locrmin_r Minimum local (per rank) range.
          * @param ntotal_ Total number of samples to generate.
         */
 	template <class RegionLayout>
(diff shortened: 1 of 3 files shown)


aaadelmann (changes requested):
Do we have a test for this function?

mohsensadr:
> Do we have a test for this function?

I added a test where we can check min/max of random variables before and after changing the boundaries.
`test/random/TestInverseTransformSamplingUpdateBounds.cpp`

I use the same density and ITS to sample position and velocity. For position inside [-2,2]^2, and velocity inside [0,2]^2. The program writes
```
dim    min(pos)    max(pos)    min(vel)    max(vel)
0 -1.9796144197e+00 1.9890737993e+00 1.7195022872e-05 1.9954672006e+00
1 -1.9997797847e+00 1.9999229952e+00 3.6506341035e-06 1.9999535412e+00
```

mohsensadr:
@matt-frey 
The current way of computing min/max of cdf for a bounded domain uses layout to find the bounds on each rank. 


`const typename RegionLayout::host_mirror_type regions = rlayout_r.gethLocalRegions();`

```
                umin_m[d] = dist_m.getCdf(regions(rank)[d].min(),d);
                umax_m[d] = dist_m.getCdf(regions(rank)[d].max(),d);
```
https://github.com/IPPL-framework/ippl/blob/4510eedee0c32f9ed3467fbda2f3116d8f36ea6d/src/Random/InverseTransformSampling.h#L70C9-L97C11


What if one wants to use inverse transform sampling without any domain decomposition? Shouldn't we have another implementation where the user only specifies the bounds of the domain, and not rlayout...

aaadelmann (approved):

matt-frey:
@mohsensadr Instead of passing the RegionLayout, we could change to just pass the domain bounds. I don't know if RegionLayout is explicitly needed. @srikrrish may know more what is possible because Sri implemented the inverse sampling.

srikrrish:
> @mohsensadr Instead of passing the RegionLayout, we could change to just pass the domain bounds. I don't know if RegionLayout is explicitly needed. @srikrrish may know more what is possible because Sri implemented the inverse sampling.

I agree.. I was thinking the same during yesterday's meeting. Apart from the bounds I don't think RegionLayout is needed anywhere else.

mohsensadr:
@matt-frey @srikrrish We essentially need bounds (min/max) of sub-domain (called local regions) corresponding to each rank. So, we can have another constructor that takes in rmin/rmax (bounds of the whole domain) as well as local_min/local_max (bounds of local regions on each rank). 

By the way, is there a way to get the local bounds without rlayout?

```
           const typename RegionLayout::host_mirror_type regions = rlayout_r.gethLocalRegions();
            
            int rank = ippl::Comm->rank();
            Vector<T, Dim> nr_m, dr_m;
            for (unsigned d = 0; d < Dim; ++d) {
                nr_m[d] = dist_m.getCdf(regions(rank)[d].max(), d) - dist_m.getCdf(regions(rank)[d].min(),d);
                dr_m[d]   = dist_m.getCdf(rmax_m[d],d) - dist_m.getCdf(rmin_m[d],d);
                umin_m[d] = dist_m.getCdf(regions(rank)[d].min(),d);
                umax_m[d] = dist_m.getCdf(regions(rank)[d].max(),d);
            }
```

src/Random/InverseTransformSampling.h
```diff
@@ -46,36 +46,161 @@ namespace ippl {
...
+
+        }
+
+        /*!
+         * @brief Constructor for InverseTransformSampling class without applying domain decomposition..
+         *
+         * @param dist_ The distribution to sample from.
+         * @param rmax_ Maximum range for sampling.
+         * @param rmin_ Minimum range for sampling.
+         * @param rlayout The region layout.
```
srikrrish, line 61: I think these comments are wrong (seems copy pasted from the other one without change). Also why it is mentioned as rmin_ and not as rmin_r in the comments?

src/Random/InverseTransformSampling.h
```diff
@@ -46,36 +46,161 @@ namespace ippl {
...
+         * @param ntotal_ Total number of samples to generate.
+        */
+	template <class RegionLayout>
+        InverseTransformSampling(Distribution &dist_r, Vector<T, Dim> &rmax_r, Vector<T, Dim> &rmin_r, Vector<T, Dim> &locrmax_r, Vector<T, Dim> &locrmin_r, size_type &ntotal_r)
+        : dist_m(dist_r),
```
srikrrish, line 66: Can this constructor itself be used for the one with rlayout also so that you specify the local minimum and maximum from rlayout from the calling side? This way we can reduce the number of constructors

srikrrish (changes requested):

src/Random/InverseTransformSampling.h
mohsensadr: I'm not following.
- What is the problem with having more constructors?
- if we remove the constructor with rlayout as argument, we have to change the constructor calls in all the alpine examples.

src/Random/InverseTransformSampling.h
mohsensadr: yes. they were copied pasted. I'll update them.

src/Random/InverseTransformSampling.h
mohsensadr: I see. Checkout 07465ec. I tried to reduce some of the redundancies.

srikrrish: Does your test still goes through?

mohsensadr: yes

srikrrish (approved):

PR #327, opened 2024-12-05
Add Custom+Partial Iteration for Particle Scatter and Gather Operations

### LINKED ISSUE #326: Change scatter/gather to allow custom/partial iteration over particles
# Goal

Integrating energy binning for the field solver (needed in OPAL-X) requires the functionality to only scatter particles inside a specific bin. Since particles might not be sorted by their bin index in the bunch, there are two aspects to consider:
1. Tell `scatter(...)` to iterate over a custom `Kokkos::range_policy(...)` passed by the user. 
2. Since Kokkos does not support arbitrary index iteration, add the possibility to use `ippl::hash_type` for defining the index.

To get the full field solution on a particle (combined from every individual bin solution), `gather(...)` needs to add the field value to the attribute, not overwrite it ("`+=`" instead of "`=`").

# Proposed Changes - `scatter`

Modify `ParticleAttrib::scatter` to accept a custom range policy as well as an `ippl::hash_type`:
```diff
template <typename T, class... Properties>
template <typename Field, class PT, typename policy_type>
void ParticleAttrib<T, Properties...>::scatter(
    Field& f, const ParticleAttrib<Vector<PT, Field::dim>, Properties...>& pp,
+   policy_type iteration_policy, hash_type hash_array) const {
    ...
+   const bool useHashView = hash_array.extent(0) > 0;
    if (useHashView && (iteration_policy.end() > hash_array.extent(0))) {
        Inform m("scatter");
        m << "Hash array was passed to scatter, but size does not match iteration policy." << endl;
        ippl::Comm->abort();
    }
    Kokkos::parallel_for(
        "ParticleAttrib::scatter", iteration_policy,
        KOKKOS_CLASS_LAMBDA(const size_t idx) {
            // map index to possible hash_map
+           size_t mapped_idx = useHashView ? hash_array(idx) : idx;

            // find nearest grid point
            vector_type l                        = (pp(mapped_idx) - origin) * invdx + 0.5;
            Vector<int, Field::dim> index        = l;
            Vector<PositionType, Field::dim> whi = l - index;
            Vector<PositionType, Field::dim> wlo = 1.0 - whi;

            Vector<size_t, Field::dim> args = index - lDom.first() + nghost;

            // scatter
+           const value_type& val = dview_m(mapped_idx);
            detail::scatterToField(std::make_index_sequence<1 << Field::dim>{}, view, wlo, whi,
                                    args, val);
        });
    ...
}
```
It uses the `hash_type` if a non-empty `hash_array` is passed and does a sanity check if the policy matches the amount of available elements. The main change is just the usage of `mapped_idx` instead of `idx`. This does not create andy branching on GPU, since `useHashView` is constant.

## Non-Class `scatter` Interface

In order to preserve old functionality, I overloaded the arguments in a second implementation:
```diff
template <typename Attrib1, typename Field, typename Attrib2, 
+         typename policy_type = Kokkos::RangePolicy<typename Field::execution_space>>
inline void scatter(const Attrib1& attrib, Field& f, const Attrib2& pp) {
    attrib.scatter(f, pp, policy_type(0, attrib.getParticleCount())); 
}

// Second implementation for custom range policy, but without index array
template <typename Attrib1, typename Field, typename Attrib2, 
+         typename policy_type = Kokkos::RangePolicy<typename Field::execution_space>>
inline void scatter(const Attrib1& attrib, Field& f, const Attrib2& pp, 
+                   policy_type iteration_policy, typename Attrib1::hash_type hash_array = {}) {
    attrib.scatter(f, pp, iteration_policy, hash_array);
}
```
It is execution space independent (same as the field) and uses the `hash_type` already defined inside the attribute.

# Proposed Changes - `gather`

Modify `ParticleAttrib::gather` to take a bool that overwrites the assignment operator:
```diff
template <typename T, class... Properties>
template <typename Field, typename P2>
void ParticleAttrib<T, Properties...>::gather(
    ...
    Kokkos::parallel_for(
        "ParticleAttrib::gather", policy_type(0, *(this->localNum_mp)),
        KOKKOS_CLASS_LAMBDA(const size_t idx) {
            ...

            // gather
            value_type gathered = detail::gatherFromField(std::make_index_sequence<1 << Field::dim>{},
                                                            view, wlo, whi, args);
+           if (addToAttribute) dview_m(idx) += gathered;
+           else                        dview_m(idx)   = gathered;
        });
    ...
}
```
It saves the gathered value and either overwrites or adds to the values of `dview_m(idx)`. A custom iteration policy in here is not necessary, since a field-per-bin still affects *all* particles.

## Non-Class `gather` Interface

As before, to preserve old functionality, a default argument is used in the Non-Class `gather`:
```diff
template <typename Attrib1, typename Field, typename Attrib2>
inline void gather(Attrib1& attrib, Field& f, const Attrib2& pp, 
+                           const bool addToAttribute = false) {
    attrib.gather(f, pp, addToAttribute);
}
```

# Testing

I added another test file `TestHashedScatter` and modified the existing `TestGather`.

aliemen:
This pull request implements the following changes to resolve issue [#326](https://github.com/IPPL-framework/ippl/issues/326):

1. **Scatter Operation Enhancements**:
   - Modified `ParticleAttrib::scatter` to accept a custom range policy and hashing array, like `ippl::hash_type`.
   - Added a check to ensure the custom policy matches the number of available elements.
   - Introduced the usage of `mapped_idx` instead of `idx` for particle scattering.
   - Overloaded the non-class `scatter` interface to preserve old functionality while allowing custom range policies.

2. **Gather**:
   - Modified `ParticleAttrib::gather` to include a boolean parameter that determines whether to add to or overwrite the attribute's values.
   - Changed the non-class `gather` interface to maintain backward compatibility.

**Testing**:
- Added a new test file `test/particle/TestHashedScatter.cpp`.
- Modified the existing `TestGather` to validate the new functionality (use "+=" instead of "+").

closes #326

### DIFF
--- a/src/Particle/ParticleAttrib.hpp
+++ b/src/Particle/ParticleAttrib.hpp
@@ -207,8 +207,11 @@ namespace ippl {
                 // gather
                 value_type gathered = detail::gatherFromField(std::make_index_sequence<1 << Field::dim>{},
                                                               view, wlo, whi, args);
-                if (addToAttribute) dview_m(idx) += gathered;
-                else                dview_m(idx)  = gathered;
+                if (addToAttribute) {
+                    dview_m(idx) += gathered;
+                } else {
+                    dview_m(idx)  = gathered;
+                }
             });
         IpplTimings::stopTimer(gatherTimer);
     }
@@ -218,13 +221,46 @@ namespace ippl {
      *
      */
 
+    /**
+     * @brief Non-class interface for scattering particle attribute data onto a field.
+     *
+     * This overload preserves legacy functionality by providing a default iteration policy.
+     * It calls the member scatter() with a default Kokkos::RangePolicy.
+     * 
+     * @note The default behaviour is to scatter all particles without any custom index mapping.
+     *
+     * @tparam Attrib1 The type of the particle attribute.
+     * @tparam Field The type of the field.
+     * @tparam Attrib2 The type of the particle position attribute.
+     * @tparam policy_type (Default: `Kokkos::RangePolicy<typename Field::execution_space>`)
+     * @param attrib The particle attribute to scatter.
+     * @param f The field onto which the data is scattered.
+     * @param pp The ParticleAttrib representing particle positions.
+     */
     template <typename Attrib1, typename Field, typename Attrib2, 
                 typename policy_type = Kokkos::RangePolicy<typename Field::execution_space>>
     inline void scatter(const Attrib1& attrib, Field& f, const Attrib2& pp) {
         attrib.scatter(f, pp, policy_type(0, attrib.getParticleCount())); 
     }
 
-    // Second implementation for custom range policy, but without index array
+    /**
+     * @brief Non-class interface for scattering with a custom iteration policy and optional index array.
+     *
+     * This overload allows the caller to specify a custom `Kokkos::range_policy` and an optional
+     * `ippl::hash_type` array. It forwards the parameters to the member scatter() function.
+     * 
+     * @note See ParticleAttrib::scatter() for more information on the custom iteration functionality.
+     *
+     * @tparam Attrib1 The type of the particle attribute.
+     * @tparam Field The type of the field.
+     * @tparam Attrib2 The type of the particle position attribute.
+     * @tparam policy_type (Default: `Kokkos::RangePolicy<typename Field::execution_space>`)
+     * @param attrib The particle attribute to scatter.
+     * @param f The field onto which the data is scattered.
+     * @param pp The ParticleAttrib representing particle positions.
+     * @param iteration_policy A custom `Kokkos::range_policy` defining the iteration range.
+     * @param hash_array An optional `ippl::hash_type` array for index mapping.
+     */
     template <typename Attrib1, typename Field, typename Attrib2, 
                 typename policy_type = Kokkos::RangePolicy<typename Field::execution_space>>
     inline void scatter(const Attrib1& attrib, Field& f, const Attrib2& pp, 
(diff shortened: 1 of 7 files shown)


src/Particle/ParticleAttrib.hpp
```diff
@@ -205,14 +218,24 @@ namespace ippl {
...
 
-    template <typename Attrib1, typename Field, typename Attrib2>
+    template <typename Attrib1, typename Field, typename Attrib2, 
+                typename policy_type = Kokkos::RangePolicy<typename Field::execution_space>>
     inline void scatter(const Attrib1& attrib, Field& f, const Attrib2& pp) {
-        attrib.scatter(f, pp);
+        attrib.scatter(f, pp, policy_type(0, attrib.getParticleCount())); 
+    }
+
+    // Second implementation for custom range policy, but without index array
```
srikrrish, line 227: A bit more explanation on this would be helpful.

srikrrish (changes requested):
@aliemen I have only very minor comments like adding/expanding documentation. Otherwise it looks good. You have tests. Is it also possible to convert them to unit tests just like the ones for scatter and gather?

aliemen:
@srikrrish I added comments/documentation for the scatter/gather functions. I added them for the implementation of the `ParticleAttrib` class as well as for the non in-class interfaces. Please let me know if you think that is too much, but I think it makes sense that way.

I haven't done this yet, but I will also try to convert the test file into two unit tests.

aliemen:
@srikrrish I now also pushed the unit tests. Since it was not a lot that was missing, I also added tests for the old/simple `scatter`/`gather`. I tested it on multiple ranks (1, 2, 4), on CPU as well as on GPU. In the following is a bit of explanation regarding the tests.

## Dimension Independence

You can simply uncomment the parameters with different dimensions under "`TestTypes`" to test for dimensions higher than 3. However, it obviously gets very memory hungry/slow when creating the fields. I tested it for a mesh size of 8 for up to 5 dimensions and this worked fine for alle tests.

At the moment, I only added 1, 2 and 3 dimensions with a grid size of 128 (such that it runs reasonably quick for the usually important dimension values). 

## What Tests Did I Add?

Here is a list of the 5 new unit tests:
1. `GatherTestReplace`. This one tests the normal (old) `gather` functionality. It initializes particles with charge `10` and a constant field with values `1.0`. After the `gather`, the charges should all be replaced with "`1.0`" (which is asserted for each particle).
2. `GatherTestIncrement`. This tests the flag `addToAttribute = true`. It initializes the charges and the field to `1.0`. After the `gather`, all charges should then have the value `2.0` (their old value plus the field value).
3. `ScatterSimpleTest`. This tests the normal (old) `scatter` functionality. It simply initializes charges with `1.0`, scatters them on a field (initialized to `0`) and then compares the total charges.
4. `ScatterCustomRangeTest`. Scatters only the first "`nLoc / 2 + rank`" particles using a custom `Kokkos::RangePolicy`, then tests the correctness like under 3).
5. `ScatterCustomHashTest`. Generates a `hash_type` of "`nLoc / 2 + rank`" random indices, scatters only those indices onto the field then tests the correctness like under 3). The random indices are chosen by shuffling an index array `[1, ..., nLoc]` and selecting the first `nLoc / 2 + rank` indices.

srikrrish (approved):

aaadelmann (approved):

PR #334, opened 2025-02-11
P3M Merge

tischwab0911:
Pull request to merge the P3M codes into the main repository.


aaadelmann:
Why did you add the plot's?  In general pdfs of results should not be in the repo.  Are these plots used in your thesis?

aliemen:
I skimmed a bit over the code (I won't have time to actually understand the solver itself until in two weeks...). But here are a view comments:
- I agree @aaadelmann, I think we should remove the `plots/...` and `python_scripts/...` folder for the merge., since they are not part of the solver itself?
- The files under `test/p3m/...` seem nice, and I think they could also easily be converted into unit tests.
- I added a view suggestions above concerning the code/file structure. I think, it might make sense to make an effort in keeping it consistent with the existing structure (mainly regarding Alpine mini apps and manager classes):
    1. Remove the `P3M3DManager`, since it is basically the same as `PicManager` (just a suggestion, there might be room for debate here).
    2. Remove `src/p3m/` and put `P3MParticleContainer` together with the high level P3M manager(s) into a `alpine/p3m/...` folder.
   
I have not yet looked into the `alpine/[...]P3M[...]` (or `src/PoissonSolvers/P3MSolver[...]`) classes yet, since (as mentioned) I don't fully understand the algorithm yet... But I suppose it is probably not necessary to look too much into it just yet for this merge?

aliemen (changes requested):
I skimmed a bit over the code (I won't have time to actually understand the solver itself until in two weeks...). But here are a view comments:
- I agree @aaadelmann, I think we should remove the `plots/...` and `python_scripts/...` folder for the merge., since they are not part of the solver itself?
- The files under `test/p3m/...` seem nice, and I think they could also easily be converted into unit tests.
- I added a view suggestions above concerning the code/file structure. I think, it might make sense to make an effort in keeping it consistent with the existing structure (mainly regarding Alpine mini apps and manager classes):
    1. Remove the `P3M3DManager`, since it is basically the same as `PicManager` (just a suggestion, there might be room for debate here).
    2. Remove `src/p3m/` and put `P3MParticleContainer` together with the high level P3M manager(s) into a `alpine/p3m/...` folder.
   
I have not yet looked into the `alpine/[...]P3M[...]` (or `src/PoissonSolvers/P3MSolver[...]`) classes yet, since (as mentioned) I don't fully understand the algorithm yet... But I suppose it is probably not necessary to look too much into it just yet for this merge?

src/P3M/P3MParticleContainer.hpp
```diff
@@ -0,0 +1,102 @@
...
+        typename Base::particle_position_type E;    // electric field at particle position
+        ippl::ParticleAttrib<double> phi;  // electric potential at particle position
+        // typename Base::particle_index_type ID;      // particle global index
+        // typename Base::particle_position_type F_sr; // short-range interaction force
+
+    private:
+        PLayout_t<T, Dim> pl_m;     // Particle layout 
+        NList_t nl_m;               // NeighborList
+        bool nlValid_m;             // true if neighbor list was initialized
+        bool *neighbors_m;
```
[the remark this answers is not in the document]
0xBachmann: this should be an array of bools. what is the best datatype to do this? is a std::vector possible or should it be a Kokkos::View?

src/Manager/P3M3DManager.h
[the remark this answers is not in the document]
0xBachmann: P3M3DManager has its particle container fixed to P3MParticleContainer instead of having a template parameter or it. does this make it reason enough to have a separate class for it?

aliemen:
> 2. Remove `src/p3m/` and put `P3MParticleContainer` together with the high level P3M manager(s) into a `alpine/p3m/...` folder.

_Update after todays IPPL Meeting_: Let's not extend "alpine", but rather create a separate directory. Also: I think it would improve clarity if we created a new "examples" directory, where we put all these mini apps in. Maybe a folder structure like the following:
``` 
examples/
├─ alpine/
│  ├─ AlpineManager.h
│  ├─ ...
├─ collision/
│  ├─ P3M[...]Manager.hpp
│  ├─ P3MParticleContainer.hpp
│  ├─ datatypes.h
│  ├─ [...].cpp 
│  ├─ ...
├─ cosmology/
│  ├─ ...
```

Also I just notices you removed `P3MParticleContainer.hpp` (which is not what I meant above). I was thinking of removing `P3M3DManager.h`, since it is basically the same as the `PicManager.h`, but with a fixed particle container type. Since `PicManager` is templated on the container type, one can easily use it together with `P3MParticleContainer.hpp`. 


@0xBachmann be carefule with the difference between a manager class and a container class ;):
- Container only saves particle information (e.g. attributes like position, charge, mass) and their layout (e.g. "where on the machine"). On a high level, they should inherit from e.g. `ParticleBase`.
- Managers manage the simulation. They hold the containers (particle/field) and all other tools (like the solver instance) and define what happens in the simulation every timestep. On a high level, they should inherit from `PicManager` for particle simulations. 

Therefore, removing the only P3MContainer is not a good idea, it should be high level part of the "P3M mini apps". Removing the code duplication introduced by the P3M3DManager, which does not need to be a high level specialization, however, should improve clarity.

examples/collisions/P3MParticleContainer.hpp
```diff
@@ -27,13 +27,13 @@ class P3MParticleContainer : public ippl::ParticleBase<ippl::ParticleSpatialOver
...
+    using particle_vector_type = typename Base::particle_position_type;
+    using particle_scalar_type = ippl::ParticleAttrib<T>;
 public:
-    ippl::ParticleAttrib<T> Q; // charge
-    typename Base::particle_position_type P; // particle velocity
-    typename Base::particle_position_type E; // electric field at particle position
-    // ippl::ParticleAttrib<double> phi;  // electric potential at particle position
-    // typename Base::particle_index_type ID;      // particle global index
-    // typename Base::particle_position_type F_sr; // short-range interaction force
+    particle_scalar_type Q; // charge
```
aaadelmann, line 33: Why is this not in the field container?

aaadelmann (changes requested):

aaadelmann (approved):

aaadelmann:
uups my accept was a mistake, need to revert

0xBachmann (approved):
I think it's now ready for the merge.

aaadelmann (approved):
Could reproduce on ALPS the results

aliemen:
The P3M code looks good to me, I'll also try to run it in a second. 
- I would add a short description to new classes (just an explanation of what it does and what it is for, similar to what was done [here](https://github.com/IPPL-framework/ippl/blob/master/src/Particle/ParticleAttrib.hpp).
- You added two new functions to `ParticleAttrib.h`, which I think makes sense. However, since this is very much core-IPPL, you should probably add documentation.

These are very general (just documenttion...). Codewise I only have a few comments on the `ParticleAttrib` changes. See below

src/Particle/ParticleSpatialOverlapLayout.hpp
```diff
@@ -0,0 +1,1123 @@
...
+            const auto min    = globalRegion[d].min();
+            const auto length = globalRegion[d].length();
+            const auto middle = min + length / 2;
+            Kokkos::parallel_for(
+                "correct positions",
+                // the copied particles are appended at the end of the attributes
+                Kokkos::RangePolicy<position_execution_space>(numLoc,
+                                                              numLoc + numBoundaryParticles),
+                KOKKOS_LAMBDA(const size_t& i) {
+                    positions(i)[d] += positions(i)[d] > middle ? -length : length;
```
aliemen, line 238: Just a comment: In this line, positions of particles that are outside of the domain of active particles are edited. It certainly works, since everything happens in one function, however, to me it is a bit sketchy. 

At the moment, I don't see a better solution except calling `setLocalNum` inside `copyAttributes`. But, again, we can leave it. 

BUT there is a worse problem:
[This](https://github.com/IPPL-framework/ippl/blob/620e6be24fdaf4afdc35f4c917edd4627e8632c2/src/Particle/ParticleSpatialOverlapLayout.hpp#L246) just updates the particles in the container - which is fine. However, [setLocalNum does not change the attribute](https://github.com/IPPL-framework/ippl/blob/620e6be24fdaf4afdc35f4c917edd4627e8632c2/src/Particle/ParticleBase.h#L145). Even worse, is it possible that all of IPPL generally **never** changes `localNum_mp` in the attribute, only `localNum_m` inside `ParticleBase` (also during communication)?. 
Maybe it is too late now and I am just confused, but if true, then this might be the reason for a lot of segmentation faults when dealing with heavily unbalanced simulations. I'll look into this again, and open an issue if this is true.

aliemen (changes requested):
To summarize: I'll approve it, if you:
- add/correct the documentation where I commented.
- use the getter instead of the hard coded localNum (also where I commented). 

(...or of course if @aaadelmann says this doesn't matter). The rest seems fine to me, my other comments are more general. The localNum problem is something I/we have to look into separately.

src/Particle/ParticleSpatialOverlapLayout.hpp
aliemen: Looks good to me.

aliemen:
@matt-frey I requested your review so that you can take a look at the ``ParticleAttrib::internalCopy` function, since you mostly wrote the class, correct?

The function works perfectly fine in my opinion, you don't need to test it. But I think it would make sense to get your opinion on its placement inside the `ParticleAttrib` class and if you think it belongs there.

0xBachmann:
I would also prefer to not have it in the base class, however in `ParticleBase` all attributes are stored as pointers to `ParticleAttribBase` and the attribute base class does not provide access to the underlying view (which is not possible as it is dependent on the deriving classes template parameters). Even with `forAllAttributes` its not possible to reproduce the same behaviour except to serialize, manipulate, deserialize.

PR #338, opened 2025-02-21
Add FEM building blocks to IPPL

s-mayani:
We are adding a Finite Element Method (FEM) framework to IPPL, with this Pull Request containing the first few building blocks on our way to having also Nédélec and Raviart Thomas elements, and higher-order FEM. Currently, only order 1 Lagrange elements are available. These building blocks of the FEM library can be found in the src/FEM directory. Unit tests have been added in unit_tests/FEM/.

 To showcase how to use these to solver PDEs, we have a FEM-based Poisson solver in src/PoissonSolvers/FEMPoissonSolver.h. The test/solver/fem/ directory contains some test problems using this solver, with different boundary conditions.
 
More information on this framework can be found in the [Bachelor Thesis of Lukas Bühler](https://amas.web.psi.ch/people/aadelmann/ETH-Accel-Lecture-1/projectscompleted/phys/bachelor_thesis_buehlluk.pdf).


src/FEM/FEMesh.h
aaadelmann, line 1: remove commented code

src/FEM/UniformCartesianFEMesh.h
aaadelmann, line 1: remove commented code

src/FEM/FiniteElementSpace.h
aaadelmann, line 1: remove commented code

aaadelmann (approved):
I made some comments on not used code (commented out) that should be removed. Otherwise nicely done with a lot of description etc.

aaadelmann:
I skimmed all files, looks good and has decent doc. I also resolved the conflicts as best as I could did however NOT compile/test

test/solver/fem/TestNonhomDirichlet_1d.cpp
aaadelmann, line 1: Describe what this test does

test/solver/fem/TestNonhomDirichlet_1dcos.cpp
aaadelmann, line 1: Describe what this test does

test/solver/fem/TestNonhomDirichlet_2d.cpp
aaadelmann, line 1: Describe what this and all other test do.

aaadelmann (approved):

src/PoissonSolvers/FEMPoissonSolver.h
```diff
@@ -0,0 +1,232 @@
...
+
+        // PCG (Preconditioned Conjugate Gradient) is the solver algorithm used
+        using PCGSolverAlgorithm_t =
+            CG<lhs_type, lhs_type, lhs_type, lhs_type, lhs_type, FieldLHS, FieldRHS>;
+
+        // FEM Space types
+        using ElementType =
+            std::conditional_t<Dim == 1, ippl::EdgeElement<Tlhs>,
+                               std::conditional_t<Dim == 2, ippl::QuadrilateralElement<Tlhs>,
+                                                  ippl::HexahedralElement<Tlhs>>>;
```
srikrrish, line 54: Is Dim == 3 missing?

src/PoissonSolvers/FEMPoissonSolver.h
```diff
@@ -0,0 +1,232 @@
...
+
+        using QuadratureType = GaussJacobiQuadrature<Tlhs, 5, ElementType>;
+
+        using LagrangeType = LagrangeSpace<Tlhs, Dim, 1, ElementType, QuadratureType, FieldLHS, FieldRHS>;
```
srikrrish, line 58: Why the orders are hardcoded here?

test/solver/fem/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
+# -----------------------------------------------------------------------------
+# test/solver/fem/CMakeLists.txt
+#
+# Integration tests for IPPL FEM solvers.
+# -----------------------------------------------------------------------------
+
+file(RELATIVE_PATH _relPath "${CMAKE_SOURCE_DIR}" "${CMAKE_CURRENT_SOURCE_DIR}")
+message(STATUS "🔧 Adding sFEM olver integration tests from: ${_relPath}")
```
srikrrish, line 8: typo in FEM solver please correct

unit_tests/FEM/CMakeLists.txt
```diff
@@ -0,0 +1,97 @@
...
+    ${Kokkos_DIR}/..
+)
+
+add_executable (EdgeElement EdgeElement.cpp)
+gtest_discover_tests(EdgeElement PROPERTIES TEST_DISCOVERY_TIMEOUT 600)
+
+target_link_libraries (
+    EdgeElement
+    ippl
+    pthread
```
srikrrish, line 20: Where do we use pthreads explicitly?

src/FEM/Elements/EdgeElement.h
```diff
@@ -0,0 +1,117 @@
...
+
+#ifndef IPPL_EDGEELEMENT_H
+#define IPPL_EDGEELEMENT_H
+
+#include "FEM/Elements/Element.h"
+
+namespace ippl {
+
+    template <typename T>
+    class EdgeElement : public Element1D<T, 2> {
```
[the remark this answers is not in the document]
srikrrish: Ok after reading through the whole PR I think I understand what this means. So this is just the basic 1D element but I got confused by the name because edge elements are also used in Nedelec. May be just clarify this in the comments.

srikrrish (changes requested):

src/FEM/Elements/EdgeElement.h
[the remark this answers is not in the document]
s-mayani: I have added a comment to clarify this, but I wonder whether I should just change the name to LineElement?

src/Meshes/UniformCartesian.h
```diff
@@ -16,37 +16,44 @@ namespace ippl {
...
+            for (unsigned int d = 0; d < Dim; d++) {
                 vertexPosition(d) = ndi[d].first() * meshSpacing_m[d] + this->origin_m(d);
+                //printf("vertexPos = %lf", vertexPosition(d));
+            }
             return vertexPosition;
         }
 
         // Vertex-vertex grid spacing of indexed cell:
-        vector_type getDeltaVertex(const NDIndex<Dim>& ndi) const {
+        KOKKOS_INLINE_FUNCTION vector_type getDeltaVertex(const NDIndex<Dim>& ndi) const override {
```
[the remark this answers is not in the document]
s-mayani: Yes this seems to be an already available function in IPPL, as it is not used anywhere in the FEM.

src/PoissonSolvers/FEMPoissonSolver.h
s-mayani: In this case Dim == 3 is the default case, which is why we don't put it

unit_tests/FEM/CMakeLists.txt
s-mayani: hmm this looks like the old cmake, not sure why you are seeing this CMakeListst.txt instead of the updated one...

src/FEM/FiniteElementSpace.h
```diff
@@ -0,0 +1,281 @@
...
+
+#include "Types/ViewTypes.h"
+
+#include "Utility/IpplException.h"
+
+#include "FEM/Elements/Element.h"
+#include "FEM/Quadrature/Quadrature.h"
+#include "Meshes/UniformCartesian.h"
+
+constexpr unsigned calculateNumElementVertices(unsigned Dim) {
```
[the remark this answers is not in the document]
s-mayani: moved it inside

src/PoissonSolvers/FEMPoissonSolver.h
s-mayani: Changed this to be a template parameter of the FEMPoissonSolver with a default value of order 1 and 5 quadrature points.

src/Field/HaloCells.h
```diff
@@ -58,6 +59,16 @@ namespace ippl {
              */
             void accumulateHalo(view_type& view, Layout_t* layout);
 
+            /*!
+             * Send halo data to internal cells for only the physical cells.
```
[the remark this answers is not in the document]
s-mayani: What I mean here is that I send only the data along the physical cells themselves, not counting the corner halo cells (which would add the value of the halo cell to another halo cell if I were to take them into account - but I don't want to modify the value of the other halo cell). I have changed the documentation a bit to reflect this, let me know if it is clearer.

src/FEM/LagrangeSpace.hpp
```diff
@@ -0,0 +1,970 @@
...
+            },
+            Kokkos::Sum<int>(upperBoundaryPoints));
+        Kokkos::fence();
+
+        int elementsPerRank = npoints - upperBoundaryPoints;
+        elementIndices      = Kokkos::View<size_t*>("i", elementsPerRank);
+        Kokkos::View<size_t> index("index");
+
+        Kokkos::parallel_for(
+            "RemoveNaNs", npoints, KOKKOS_CLASS_LAMBDA(const int i) {
```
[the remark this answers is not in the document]
s-mayani: This is because in order to distribute the elements among MPI ranks, I access the local domain of the MPI rank and fill out the elementIndices view with the element index corresponding to each point in the local domain. This element index is always computed taking the point as the first vertex of the element (if you think in 1D, this means that DOF 0 corresponds to element 0, DOF 1 to element 1, etc. i.e. the edge element always lies to the right of the vertex point). However, this procedure would create a fictitious element outside of the domain boundary if I apply it to a boundary point. Which is why I tag those boundary points as invalid in the previous Kokkos loop, which is also a parallel reduce because I need to be able to set the size of the elementIndices view (which will be npoints - boundaryPoints). Then I assign the correct values to the elementIndices view in this parallel_for, by skipping the invalidated boundary points. I added some comments in the code itself to make it clearer.

src/FEM/Elements/HexahedralElement.hpp
```diff
@@ -0,0 +1,103 @@
+namespace ippl {
+    template <typename T>
+    KOKKOS_FUNCTION typename HexahedralElement<T>::vertex_points_t
+    HexahedralElement<T>::getLocalVertices() const {
+        HexahedralElement::vertex_points_t vertices = {
+            {0.0, 0.0, 0.0}, {1.0, 0.0, 0.0}, {1.0, 1.0, 0.0}, {0.0, 1.0, 0.0},
```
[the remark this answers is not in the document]
s-mayani: Yes, this follows the indexing in FEMSTER. Maybe I will just put a comment referencing Lukas' thesis where the local vertex numbering is specified for each element type.

unit_tests/FEM/CMakeLists.txt
srikrrish: Yeah it does say outdated.

srikrrish (approved):

PR #341, opened 2025-03-29
Modernized CMake System.

JoMee:
This PR modernizes the CMake system by:

- Splitting it up into a sensible structure. 
- Adding symbolic targets to link against.
- Allows for exporting IPPL as a library using the IPPL_DYL flag.
- Adding unit and integration tests to be discovered by CTest.
- Allows address sanitizers to run (if linked against an implementation of MPI which supports this such as MPICH).

This new structure was tested locally and on JEDI so far, but we have decided to "push and pray" and fix things in the new cmake structure if issues come up.


aaadelmann:
All variables in **IpplVersions.h.in** are not defined

JoMee:
What exactly do you mean by the variables are not defined? Are you referring to the README? Should I add them there as well?

JoMee:
Clang should definitely support the sanitizer, what often doesn't work is using the sanitizer while linking against the openMPI implementation of MPI. If that is the issue, you can try to link against mpich: https://www.mpich.org/

aaadelmann:
Yes and then mpich is not available or compatible with Clang. We had a lot of problem with this hence, sometimes less is more this is why I have expluded for the moment Clang.

aaadelmann (approved):

JoMee:
I compiled it with clang now and I see what's the issue. clang does not link to the standard library (libstdc++). If you compile with clang++ that does work (I do get it to compile and run with that) or we can also link our targets manually against libstdc++ then clang should work with the sanitizer. But compiling with "-DCMAKE_CXX_COMPILER=clang++" works for me. I guess since this is a c++ project it does make sense to use clang++?

matt-frey (commented):
Is there a reason for these emojis? What happens when these are not supported by the shell?

JoMee:
The idea was to somehow visually distinguish between different kinds of messages, those are status messages and should jump out more easily than some info or warnings. The emojis are UTF-8 encoded and I would say that it is fair to assume UTF-8 is supported. If emojis can be displayed is a different question but this is not dependent on your shell (the shell will support it), this is dependent on your terminal and locally installed fonts.

JoMee:
I did compile it in debug mode with clang (not clang++) and that seems to be working with sanitizer as well. I just had to provide it with the right cmake flags, namely "-DCMAKE_PREFIX_PATH" and "-DCMAKE_C_COMPILER" and make sure i have the mpich in my path and also linker library path (for example by setting "export PATH=/path/to/mpich-install/bin:$PATH" and "export LD_LIBRARY_PATH=/path/to/mpich-install/lib:$LD_LIBRARY_PATH").

Maybe if that involves too many steps in terms of knowing how to deal with different installations of mpi we can make the option of sanitizers with clang separate. But it would be useful to my testing system if that should also test clang builds.

aaadelmann:
please try on Merlin if you have not already.

JoMee:
If I see this correctly, I don't have access to Merlin anymore. But I don't see of a reason the sanitizer wouldn't work on there. And is there even a reason to compile IPPL on merlin with debug + sanitizer?

aaadelmann:
Jonas your account at PSI is still enabled. Did you had already access to merlin?

JoMee:
Somehow, access to my VPN isn't working anymore, not sure about merlin.

s-mayani (commented):
typo in tests/CMakeLists.txt

s-mayani (approved):

PR #347, opened 2025-04-26
remove global constant PI from datatypes.h

aaadelmann:
remove global constant PI from datatypes.h


mohsensadr:
With this fix (removing global constant PI from datatypes.h) and the updated opalx's cmakelist and command, opalx successfully compiles on Merlin.

rammann (approved):

mohsensadr (approved):

mohsensadr:
Although pi could be defined in AlpineManger instead to avoid repetition, the current approach is also ok from my perspective.

PR #348, opened 2025-04-27
sanitizer option

aaadelmann:
allow the user to select sanitizer

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,7 @@ option(IPPL_ENABLE_TESTS "Build integration tests in test/ directory" OFF)
 option(IPPL_DYL "Build IPPL as a shared library (ON) or static library (OFF)" OFF)
 option(IPPL_ENABLE_COVERAGE "Enable code coverage" OFF)
 option(IPPL_ENABLE_NSYS_PROFILER "Enable Nvidia Nsys Profiler" OFF)
+option(IPPL_ENABLE_SANITIZER "Enable sanitizer(s)" OFF)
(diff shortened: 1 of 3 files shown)


rammann (approved):

mohsensadr (commented):
If I recall correctly, the combination of sanitizer and Cuda does not work. Shouldn't the cmake make sure that in such a scenario (Sanitizer+Cuda), the sanitizer is set to OFF, even if the user by accident has enabled it?

aaadelmann:
On the host the sanitizer is working but not on the device. So it is not clear to me how we will handle this. I guess we will see in the future.

mohsensadr:
Is it safe to say that sanitizer can be only turned on when `DIPPL_PLATFORMS` is either "serial" or "openmp", which implies execution space is host? This should exclude device related ones, like "cuda" and "HIP;OPENMP" (I'm guessing exec. space of "HIP;OPENMP" is device).

aaadelmann:
"serial" or "openmp" with GCC and (if I am not mistaken CLANG) is save a combination that should work. Not sure if we can rule all the other out.

s-mayani (approved):

mohsensadr:
> "serial" or "openmp" with GCC and (if I am not mistaken CLANG) is save a combination that should work. Not sure if we can rule all the other out.

I see. I'm also not completely sure of all the combinations that sanitizer would work. Maybe it is a good idea to leave it to the user then.

mohsensadr (approved):

aaadelmann:
Updated README.md

PR #352, opened 2025-05-06
remove the Cuda/HIP backend macros for the HostSpace FFT definition

s-mayani:
Unit tests are not compiling because of a change in FFT.h when adding HIP/rocm support. 

This is because the Kokkos::HostSpace FFT needs to be defined even if we have Cuda or HIP (since these are on Kokkos::CudaSpace or Kokkos::HIPSpace).

I do not know if this breaks anything on AMD (please let me know Andreas). 
If it does, then I think we need to somehow add these macros to the unit tests instead.

### DIFF
--- a/src/FFT/FFT.h
+++ b/src/FFT/FFT.h
@@ -123,7 +123,7 @@ namespace ippl {
 #endif
 #endif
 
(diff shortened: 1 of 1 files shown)


biddisco:
This fixes the build error for me too

aaadelmann (approved):
did not break LUMI

JoMee (approved):

PR #358, opened 2025-05-27
CMake tweaks and cleanup to help with alps uenv builds

biddisco:


CMakeUserPresets.json
```diff
@@ -0,0 +1,64 @@
+{
```
aaadelmann, line 1: Would be nice to have a header describing the purpose of this file

aaadelmann (changes requested):

JoMee:
I like those changes a lot, nice!

CMakeLists.txt
```diff
@@ -1,32 +1,88 @@
...
-option(USE_ALTERNATIVE_VARIANT "Use modified variant implementation (required for CUDA 12.2 + GCC 12.3.0)" OFF)
+message(STATUS
+    "📦 Configuring IPPL Version: ${IPPL_VERSION_MAJOR}.${IPPL_VERSION_MINOR} : \"${IPPL_VERSION_NAME}\""
+)
 
+# ------------------------------------------------------------------------------
+# Module path
+# ------------------------------------------------------------------------------
+list(PREPEND CMAKE_MODULE_PATH
+  "${PROJECT_SOURCE_DIR}/cmake" "${PROJECT_SOURCE_DIR}/CMakeModules")
```
JoMee, line 27: Do we want both a cmake and a CMakeModules folder? It might be more clean to have a folder-hierarchy all inside the cmake folder? Or what would the arguments be for having both?

CMakeUserPresets.json
biddisco: I'm not sure what the best thing to do there is, because json doesn't support comments. I'm not sure if anything much breaks of they are added, but since the CMakeUserPresets is a "standard" cmake file, I kind of assumed it was obvious what it was for. I was intending to add some stuff to the readme so perhaps that's the best place to put a description since the file defines some default setup cmake configurations.

CMakeLists.txt
biddisco: I think that I just didn't notice there was the CMakeModules there - it doesn't exist on my tree, so is it a left over from somewhere, or should it be present. Probably I just added the cmake and missed the other, we can/should certainly remove one

JoMee: From what I can see there is no official recommendation what name one should choose here but I've looked at some more modern projects and they mainly seem to be using "cmake" so I'd suggest using just that one?

CMakeUserPresets.json
aaadelmann: ok mabee such things need to go to a Read.me

aaadelmann:
Please pull first the master, I have a feeling you have 101 conflicts.

PR #367, opened 2025-06-30
Fix compilation issue clang

s-mayani:
Fixes "non-type template argument is not a constant expression" error which appears in FEM for some compilers (clang), and causes the master to not compile on Lumi.


src/FEM/Quadrature/Quadrature.hpp
```diff
@@ -19,10 +19,10 @@ namespace ippl {
     Quadrature<T, NumNodes1D, ElementType>::getWeightsForRefElement() const {
         Vector<T, NumNodes1D> w = this->getWeights1D(0.0, 1.0);
 
-        Vector<T, this->numElementNodes> tensor_prod_w;
+        Vector<T, std::remove_reference_t<decltype(*this)>::numElementNodes> tensor_prod_w;
```
aaadelmann, line 22: will that run on a device?

aaadelmann (approved):

PR #375, opened 2025-07-20
Clean testing system (fix)

JoMee:
Those are the changes I only pushed to the gitlab but not yet to the github.

### DIFF
--- a/.github/workflows/mirror.yml
+++ b/.github/workflows/mirror.yml
@@ -14,7 +14,7 @@ jobs:
   check_permission:
     runs-on: ubuntu-latest
     if: >-
-        (github.repository_owner == 'JoMee') &&
+        (github.repository_owner == 'aaadelmann') &&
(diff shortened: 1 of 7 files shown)


.github/workflows/mirror.yml
```diff
@@ -14,7 +14,7 @@ jobs:
   check_permission:
     runs-on: ubuntu-latest
     if: >-
-        (github.repository_owner == 'JoMee') &&
+        (github.repository_owner == 'aaadelmann') &&
```
srikrrish, line 17: What implications does this have when we specify a particular person as the owner?

aaadelmann (approved):

.github/workflows/mirror.yml
srikrrish: Ok.

srikrrish (approved):

PR #387, opened 2025-08-04
Empty domains in Field Layout not permitted

### LINKED ISSUE #386: FieldLayout: more ranks than local domains
Given a global domain, ippl should not allow to create a FieldLayout, if the domain can be split in to less local domains than we have ranks.
This is needed to make sure that we do not end up having ranks that have an empty local domain.

The constructor of the FieldLayout class currently still allows this to happen, the error is only caught later in the function:
void HaloCells<T,Dim,ViewArgs...>::accumulateHalo_noghost(view_type& view, Layout_t* layout, int nghost), located in ippl/scr/Field/HaloCells.hpp.

anekslen:
Do not allow the creation of fields, that have less possible local domains than ranks.

Split const NDIndex_t& getLocalNDIndex(int rank = -1) const; in to two functions, one without argument, that returns the domain of the current rank and one with a rank argument, that returns the domain of the rank that is asked for.
This solution prevents the case, where a random negative number is given for the rank. Now this prevented by an assert, before it just returned the domain of the current rank.

Closes #386

### DIFF
--- a/ci/tests.yml
+++ b/ci/tests.yml
@@ -27,14 +27,19 @@ variables:
     value: "false"
     options: ["true", "false"]
     description: "Run the compilation test for the Jureca supercomputer."
+  RUN_COMPILE_JURECA_AMPERE80:
+    value: "true"
+    options: ["true", "false"]
+    description: "Run the compilation test for the Jureca ampere80 architecture."
 
 include:
   - local: 'ci/tests/serial_unit.yml'
   - local: 'ci/generate/generate_report.yml'
-  - local: 'ci/tests/compile_gcc_debug.yml'
   - local: 'ci/tests/compile_gcc_release.yml'
   - local: 'ci/tests/compile_clang_debug.yml'
   - local: 'ci/tests/compile_jureca_release.yml'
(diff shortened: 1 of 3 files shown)


src/FieldLayout/FieldLayout.hpp
```diff
@@ -86,17 +86,17 @@ namespace ippl {
...
         long totparelems = 1;
         for (unsigned d = 0; d < Dim; ++d) {
-            totparelems *= domain[d].length();
+            totparelems *= isParallelDim_m[d] ? domain[d].length() : 1;
         }
 
         if (totparelems < nRanks) {
-            nRanks = totparelems;
+            throw std::runtime_error("FieldLayout:initialize: domain can only be partitioned in to "
+                + std::to_string(totparelems) + " local domains, but there are " + std::to_string(nRanks) + " ranks, increase the number of possible local domains.");
```
s-mayani, line 99: I would change the last part of the text to rather be "decrease the number of ranks or increase the domain." What do you think?

anekslen: Sounds good to me, changed it.

s-mayani (approved):

PR #431, opened 2025-10-15
no toss_cookies on device

aaadelmann:


aaadelmann:
I will check the name of the macro, thanks for spotting that

PR #437, opened 2025-10-28
Interpolate Current

JoMee:
This PR adds the segment splitting as well as the current state of the interpolation scheme.

### DIFF
--- a/src/FEM/GridPathSegmenter.hpp
+++ b/src/FEM/GridPathSegmenter.hpp
@@ -1,0 +1,123 @@
+#include <array>
+#include <cmath>
+#include <type_traits>
+
+namespace ippl {
+
+template<int I, int End, class F>
+KOKKOS_INLINE_FUNCTION
+constexpr void static_for(F&& f) {
+  if constexpr (I < End) {
+    f(std::integral_constant<int, I>{});
+    static_for<I+1, End>(std::forward<F>(f));
+  }
+}
+
+template<typename T>
+KOKKOS_INLINE_FUNCTION
+void sort2(T& a, T& b) { if (a > b) { T t=a; a=b; b=t; } }
+
+template<typename T>
+KOKKOS_INLINE_FUNCTION
+void sort3(T& a, T& b, T& c) { sort2(a,b); sort2(b,c); sort2(a,b); }
+
+template<unsigned Dim, typename T>
+KOKKOS_INLINE_FUNCTION
+std::array<T,Dim> lerp_point(const std::array<T,Dim>& A,
+                             const std::array<T,Dim>& B, T t) {
+  std::array<T,Dim> out{}; 
+  for (unsigned a=0; a<Dim; ++a) out[a] = A[a] + (B[a]-A[a]) * t;
+  return out;
+}
+
+// ---------- per-axis cut times (for DefaultCellCrossingRule) ----------
+template<unsigned Dim, typename T>
+struct CutTimes { std::array<T,Dim> t; };
+
+template<unsigned Dim, typename T>
+KOKKOS_INLINE_FUNCTION
+CutTimes<Dim,T> compute_axis_cuts_default(
+  const std::array<T,Dim>& A,
+  const std::array<T,Dim>& B,
+  const std::array<T,Dim>& origin,
+  const std::array<T,Dim>& h)
+{
+  CutTimes<Dim,T> cuts;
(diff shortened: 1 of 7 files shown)


src/FEM/GridPathSegmenter.hpp
```diff
@@ -0,0 +1,123 @@
+#include <array>
+#include <cmath>
+#include <type_traits>
+
```
aaadelmann: Same comments than for the .h file

src/FEM/GridPathSegmenter.hpp
```diff
@@ -0,0 +1,123 @@
...
+  return cuts;
+}
+
+// ---------- endpoints builder: [0, t_sorted..., 1] (keeps duplicates) ----------
+template<unsigned Dim, typename T>
+KOKKOS_INLINE_FUNCTION
+void make_endpoints_fixed(const CutTimes<Dim,T>& cuts,
+                          std::array<T,Dim+2>& Tcuts /*out*/)
+{
+  T t0 = cuts.t[0];
```
mohsensadr: what if `Dim==1`? If that's never the case, then consider having a check. Because with Dim=1, I think the else in line 82 would be out of bound.

src/FEM/ProjectCurrent.hpp
```diff
@@ -0,0 +1,42 @@
...
+          typename policy_type = Kokkos::RangePolicy<>>
+inline void assemble_current_whitney1(const Mesh& mesh,
+                                      const ChargeAttrib& q_attrib,
+                                      const PosAttrib& X0,
+                                      const PosAttrib& X1,
+                                      FEMVector& fem_vector,
+                                      const NedelecSpace& space,
+                                      policy_type iteration_policy)
+{
+  using T = Mesh::value_type;
```
mohsensadr: Shouldn't it be?

`using T = typename Mesh::value_type;`

mohsensadr (changes requested):

src/FEM/GridPathSegmenter.hpp
```diff
@@ -0,0 +1,123 @@
...
+ippl::Vector<T,Dim> lerp_point(const ippl::Vector<T,Dim>& A,
+                             const ippl::Vector<T,Dim>& B, T t) {
+  ippl::Vector<T,Dim> out{}; 
+  for (unsigned a=0; a<Dim; ++a) out[a] = A[a] + (B[a]-A[a]) * t;
+  return out;
+}
+
+// ---------- per-axis cut times (for DefaultCellCrossingRule) ----------
+template<unsigned Dim, typename T>
+struct CutTimes { std::array<T,Dim> t; };
```
s-mayani: Is this array only going to be accessed on host? I don't think std::array is accessible within Kokkos kernels on device. One would need to use: https://kokkos.org/kokkos-core-wiki/API/core/stl-compat/Array.html, or ippl Vectors.

src/FEM/GridPathSegmenter.hpp
```diff
@@ -0,0 +1,123 @@
...
+  const ippl::Vector<T,Dim>& h)
+{
+  CutTimes<Dim,T> cuts;
+  for (unsigned a=0; a<Dim; ++a) cuts.t[a] = (T)2; // sentinel (>1)
+
+  const T eps1 = (T)1e-12, one = (T)1;
+
+  auto axis_cut = [&](auto Ax) {
+    constexpr int a = Ax;
+    const T d   = B[a] - A[a];
```
s-mayani: Maybe rename d to dist to avoid confusion with d = Dim (which is the case in many places in ippl)?

src/FEM/GridPathSegmenter.hpp
```diff
@@ -0,0 +1,123 @@
...
+
+  static_for<0,Dim>(axis_cut);
+  return cuts;
+}
+
+// ---------- endpoints builder: [0, t_sorted..., 1] (keeps duplicates) ----------
+template<unsigned Dim, typename T>
+KOKKOS_INLINE_FUNCTION
+void make_endpoints_fixed(const CutTimes<Dim,T>& cuts,
+                          std::array<T,Dim+2>& Tcuts /*out*/)
```
s-mayani: same comment as before for std::array

s-mayani (commented):
Have reviewed the grid path segmenter, and the preliminary structure of the current projecter looks good.

PR #438, opened 2025-10-29
Fix interpolation on GPU

s-mayani:
FEMInterpolate.hpp does not work on GPU. Error: CUDA ILLEGAL INSTRUCTION.

It seems that this is happening due accessing functions of the FEM space inside the Kokkos loop, which cannot be done as the FEM space contains host functions too, so trying to access its functions on device results in errors, as the FEMInterpolate is not within the FEM space class so cannot put the loop inside the class and use KOKKOS_CLASS_LAMBDA.

To fix this, there is now a possibility to create a lightweight device copy of a FEM space through the function space.getDeviceMirror(), which returns a struct containing all the relevant members and functions of the LagrangeSpace class which are needed in FEMInterpolate, and which contains stuff which can be fully on device (only KOKKOS_FUNCTIONS). The signature of the FEMInterpolate functions have not changed, but inside the FEMInterpolate functions we get a device mirror of the space (so a DeviceStruct) before going into the Kokkos loop, and then access the functions of this one.

Minor changes: changed the `locate_element_nd_and_xi` function to not take the mesh but only the quantities it needs (mesh spacing and origin).

### DIFF
--- a/cmake/FailingTests.cmake
+++ b/cmake/FailingTests.cmake
@@ -10,8 +10,6 @@
 
 if(BUILD_TESTING AND IPPL_SKIP_FAILING_TESTS)
   set(IPPL_DISABLED_TEST_LIST
-      AssembleRHS
-      InterpolateDiracs
       ParticleSendRecv
       ORB
       PIC


srikrrish:
> Adding a KOKKOS_INLINE_FUNCTION constructor did not help. Should I try to go the helper class way or are we fine with this as it is?

If it is relatively easy and not make take much time it can be done. Otherwise I would suggest to defer it to prioritize other efforts for the paper.

PR #442, opened 2025-11-02
Cmake install + script generation improvements 

biddisco:
1) Make install rules consistent with usual practice - if project name is "IPPL", then use IPPLConfig.cmake/targets/version/etc naming with the same capitalization.
also make the IPPL:: alis target namespace the same
Fix a problem with In tree compiled Heffte when make install missed a target.
Allow IPPL to build kokkos or Heffte from source and have them found correctly by dependent project such as OpalX when they Find_package(IPPL  

2) Improve scripts for alps/lumi/etc machines so that different launch scripts and wrappers for env vars etc are used. These can now be used as examples for anyone new to the project to work from when 

3) Turn off coloured messages when using ccmake as it produces escaped chars in the output which is unsightly

4) Fix a couple of minor cmake typos and ,missing vars

### DIFF
--- a/CMakeUserPresets.json
+++ b/CMakeUserPresets.json
@@ -23,9 +23,9 @@
         "BUILD_SHARED_LIBS": "ON",
         "CMAKE_BUILD_TYPE": "Release",
         "CMAKE_EXPORT_COMPILE_COMMANDS": "ON",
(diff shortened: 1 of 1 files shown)


aaadelmann:
CI is failing on this one

aaadelmann (approved):

PR #454, opened 2025-12-03
Change --prefix to --preset

aliemen:
Cmake preset and prefix are not the same. To load a preset, one has to use e.g. `--preset=default`, not `--prefix=detault`. Prefix is used in the `cmake --install` process to change where the files are copied during installation (not needed here), while Presets are what we want to use here.


biddisco (approved):
good catch.

PR #455, opened 2025-12-15
Buffer cleanup

biddisco:
This PR contains cleanups to buffer management that should improve default performance
1) Make buffer manager a singleton so that creation of communicator objects share the same buffer handler and reuse the same caches of buffers
2) Fix a missing freeAllBuffers is particleUpdate
3) replace the default kokkos views that for the storage of the buffers with unmanaged kokkos views that have stronger alignment (in this PR 1024 bytes) than the default kokkos views (normally 64 bytes)
4) make the default buffer handler the default, rather than the logging one

This PR also adds support for logging using the spdlog library and the fmt library for clean messages, this can be enabled or disabled at compile time using IPPL_LOG_LEVEL off, trace, debug, warn, error, critical etc etc by default it is off, to use it, one needs to find_package spdlog, fmt which are now provided in uenvs for alps

**Actions Before Merge**
The _Buffer Cleanup_ will make IPPL better scalable and in 
addition make the code cleaner plus introduces some login capabilities. This commit will change code at  
the heart of IPPL. Hence, we have to be super careful and make sure important mini-apps and OPALX are 
working.  I propose the following tests that we need to execute on various architectures and most important
document the results. 

Daint
cmake: *-DIPPL_ENABLE_ALIGNED_COMMUNICATION_BUFFERS=ON*
export MPICH_GPU_IPC_ENABLED=0
export MPICH_GPU_IPC_CACHE_MAX_SIZE=256

| Test        | Solver | Machine   | Problem Size   | Ranks      | Notes/Flags  | Link to result | Who?	  |
|-------------|--------|-----------|--------------  |----------- |--------------|----------------|--------|
|LandauDamping| FFT    | Daint GPU | 1024**3, 8 PPC | 32-2048 | IPC no |                | Sonali   |
|LandauDamping| FFT    | Daint CPU |  512**3, 8 PPC | 32-512 |              |                | Sonali   |
|LandauDamping| FFT    | LUMI-G    | 1024**3, 8 PPC | 32-2048 | IPC no |                | Sonali |
|LandauDamping| FFT    | Beverin   | 1024**3, 8 PPC | 32-512 | IPC no |                | Sonali   |
|LandauDamping| FEM    | Daint GPU | 1024**3, 8 PPC | 32-2048 | IPC no |                | Sonali   |
|LandauDamping| FEM    | Daint CPU |  512**3, 8 PPC | 32-512 |              |                | Sonali   |
|LandauDamping| FEM    | LUMI-G    | 1024**3, 8 PPC | 32-2048 | IPC no |                | Sonali |
|LandauDamping| FEM    | Beverin   | 1024**3, 8 PPC | 32-512 | IPC no |                | Sonali   |
|LandauDamping| PCG    | Daint GPU | 1024**3, 8 PPC | 32-2048 | IPC no |                | Sonali   |
|LandauDamping| PCG    | Daint CPU |  512**3, 8 PPC | 32-512 |              |                | Sonali   |
|LandauDamping| PCG    | LUMI-G    | 1024**3, 8 PPC | 32-2048 | IPC no |                | Sonali |
|LandauDamping| PCG    | Beverin   | 1024**3, 8 PPC | 32-512 | IPC no |                | Sonali   |
|OPALX Drift-1| FFT    | LUMI-G    | 1024**3, 8 PPC | 32-2048 |              |                | AA     |

### DIFF
--- a/INSTALLATION.md
+++ b/INSTALLATION.md
@@ -122,7 +122,6 @@ cmake .. \
       -DBUILD_SHARED_LIBS=ON \
       -DCMAKE_HIP_ARCHITECTURES=gfx90a \
       -DCMAKE_HIP_FLAGS=--offload-arch=gfx90a \
-      -DKokkos_ENABLE_DEBUG_BOUNDS_CHECK=ON \
       -DKokkos_ENABLE_DEBUG=OFF \
       -DKokkos_ARCH_ZEN3=ON \
       -DKokkos_ARCH_AMD_GFX90A=ON \
--- a/cmake/AddIpplTest.cmake
+++ b/cmake/AddIpplTest.cmake
@@ -65,7 +65,7 @@ function(add_ippl_test TEST_NAME)
   # Link libraries (ippl exports includes/flags/deps; pick your GTest flavor)
   if(TEST_USE_GTEST_MAIN)
     target_link_libraries(${TEST_NAME} PRIVATE IPPL::ippl GTest::gtest_main)
-  else()
+  elseif(NOT TEST_INTEGRATION)
     target_link_libraries(${TEST_NAME} PRIVATE IPPL::ippl GTest::gtest)
   endif()
 
--- a/src/Utility/IpplTimings.cpp
+++ b/src/Utility/IpplTimings.cpp
@@ -128,7 +128,7 @@ void Timing::print() {
     Inform msg("Timings");
     msg << level1 << "---------------------------------------------";
     msg << "\n";
-    msg << "     Timing results for " << ippl::Comm->size() << " nodes:"
+    msg << "     Timing results for " << ippl::Comm->size() << " rank(s):"
         << "\n";
     msg << "---------------------------------------------";
     msg << "\n";
--- a/test/maxwell/README.md
+++ b/test/maxwell/README.md
@@ -11,7 +11,7 @@ This directory contains testcases for both the Standard and Non-Standard FDTD so
 
 1. **Build the executables** (in existing IPPL build directory):
    ```bash
-   cd ippl/build/test/solver/
+   cd ippl/build/test/maxwell/
    make TestStandardFDTDSolver
    make TestNonStandardFDTDSolver
    ```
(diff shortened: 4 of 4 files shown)


aaadelmann:
with GPU off you mean: -D Heffte_DISABLE_GPU_AWARE_MPI=OFF ?

aaadelmann:
The last commit seams to break the compilation: https://ippl-bc4558.pages.jsc.fz-juelich.de/365538/allure-report/index.html#suites/0a150a27af0a5eb1faedcb6100255f56/782113b35712612a/

srikrrish (approved):

s-mayani:
Here the scalings from the paper repeated with this branch, with a comparison for different problem sizes and machines.

- 512^3 and 8 particles per cell, on ALPS and LUMI:
[TOTAL_scalings_512_paper_alps.pdf](https://github.com/user-attachments/files/25790872/TOTAL_scalings_512_paper_alps.pdf)
[TOTAL_scalings_512_paper_lumi.pdf](https://github.com/user-attachments/files/25790873/TOTAL_scalings_512_paper_lumi.pdf)


- 1024^3 and 8 particles per cell, on ALPS and LUMI:
[TOTAL_scalings_1024_paper_alps.pdf](https://github.com/user-attachments/files/25790874/TOTAL_scalings_1024_paper_alps.pdf)
[TOTAL_scalings_1024_paper_lumi.pdf](https://github.com/user-attachments/files/25790875/TOTAL_scalings_1024_paper_lumi.pdf)

On LUMI, the bigger problem size crashed for the FFT solver due to some memory issue, which is why we only have a grid-point at 256 GPUs. The reason for this is unknown, but is happening only after the update of LUMI (which is why we do have data points from the paper scalings, since these were done before the update). 

Conclusions: On Alps, the buffer cleanup branch produces slightly better results for the FFT solver. For PCG and FEM, the scalings are very similar to the paper results. On Lumi, the results stay similar to the paper, maybe slightly slower (but within computational fluctuations).

src/Communicate/Archive.hpp
```diff
@@ -4,99 +4,116 @@
...
+#include "Utility/Logging.h"
+
 #include "Archive.h"
 
 namespace ippl {
     namespace detail {
 
-        template <class... Properties>
-        Archive<Properties...>::Archive(size_type size)
+        template <typename BufferType>
```
aaadelmann: This is a significant architectural change,  the buffer type is now explicitly passed as a template argument. This needs a justification.

aaadelmann (changes requested):

PR #456, opened 2025-12-16
Move not all but only the latest few entries to be published

jakob-fritz:
Only the newest few entries are moved to the folder, that is later published on the website


jakob-fritz:
Thank you @srikrrish for asking. The Jobs timed out, as the SLURM-Jobs did not run in time but needed to wait in the queue. After the increasing the timeout , rerunning the pipeline should work.

PR #459, opened 2026-01-09
Add FEM solver to Alpine

s-mayani:
Adds the FEMPoissonSolver as an option in Alpine.

Major changes:
- FEM and FEM_PRECON added as solver types in Alpine, and corresponding initialisation functions. 
- Interpolation schemes for FEM (scatterFEM and gatherFEM) which use the FEMInterpolate.h functions.
- Re-use fields in both the CG linear solver and the FEM solver. 
- Re-using a `resultField` in the LagrangeSpace has side effects in the PCG; to avoid this, we add deep copies in the PCG, since many different operators are being called in the algorithm and rewriting over the same resultField.
- The scatterFEM had a missing mass matrix --> currently fixed using an approximation, needs to be properly implemented in a future PR.

### DIFF
--- a/alpine/LandauDampingManager.h
+++ b/alpine/LandauDampingManager.h
@@ -419,6 +419,7 @@ public:
         fieldEnergy = fieldEnergy / this->totalP_m;
 
         if (ippl::Comm->rank() == 0) {
+            std::filesystem::create_directory("data");
             std::stringstream fname;
             fname << "data/FieldLandau_";
             fname << ippl::Comm->size();


alpine/AlpineManager.h
```diff
@@ -108,38 +109,96 @@ class AlpineManager : public ippl::PicManager<T, Dim, ParticleContainer<T, Dim>,
...
+        Field_t<Dim>* rho                        = &this->fcontainer_m->getRho();
+        double Q                                 = Q_m;
+        size_type localParticles                 = this->pcontainer_m->getLocalNum();
+
+        using exec_space = typename Kokkos::View<const size_t*>::execution_space;
+        using policy_type = Kokkos::RangePolicy<exec_space>;
+        policy_type iteration_policy(0, localParticles);
+
+        auto* solver = dynamic_cast<FieldSolver_t*>(this->fsolver_m.get());
+        auto& space = solver->getSpace();
```
aaadelmann: say what space is

s-mayani:
The normal PoissonCG is not working anymore with the deepCopy changes made for FEM. Changed PR into draft and looking into it.

test/field/TestInner.cpp
```diff
@@ -0,0 +1,84 @@
+// Tests the Laplacian on a scalar field
```
aaadelmann: more details needed, how is the test conceived, what are the expected results

test/solver/TestCG.cpp
```diff
@@ -0,0 +1,172 @@
+
```
aaadelmann: see comment in TestInner.cpp

PR #462, opened 2026-02-02
P3m benchmark

0xBachmann:
Fixed a few small bugs while running benchmarks


aaadelmann:
Important that we get the same results on all architectures. Next is performance.

PR #489, opened 2026-03-25
Fix load balancing with FEM

s-mayani:
This closes #485

The issue is that after load balancing, the FEM space used by the solver does not update its partitioning of the element indices among ranks according to the new domain decomposition, and the member variable resultField_m also needs to update its layout. 

This issue was correctly identified by the AI tool Hugo in #487, but not properly fixed (fixed the symptom instead of the cause). 

To fix this properly, a new `updateLayout` member function is introduced in the LagrangeSpace, which is called by the LoadBalancer in alpine if using the FEM solver, in the same way that the `updateLayout` of the fields works in the LoadBalancer. This means that every time load balancing is done and the domain decomposition changes, the LagrangeSpace takes the new domain decomposition into account and repartitions the elements too.

### DIFF
--- a/alpine/LoadBalancer.hpp
+++ b/alpine/LoadBalancer.hpp
@@ -59,17 +59,22 @@ public:
         (*E_m).updateLayout(*fl);
         (*rho_m).updateLayout(*fl);
 
-        if (fs_m->getStype() == "CG" || fs_m->getStype() == "PCG") {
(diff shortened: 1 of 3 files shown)


s-mayani:
I would suggest merging this and opening an issue for the large GPU count issue so I can look into it later.

aaadelmann (approved):

PR #494, opened 2026-04-21
Implement `shiftedGreensFunction` in `FFTOpenPoissonSolver` + symmetric Dirichlet test

aliemen:
This pull request adds support for convolving with a shifted Green's function in the FFT open boundary Poisson solver, enabling the method of images for Dirichlet boundary conditions needed in OPALX. It introduces both the implementation and a test for this feature.

## New shifted Green's function support

- Added the `shiftedGreensFunction(const Vector<double, Dim>& shift)` method to `FFTOpenPoissonSolver`, which fills the cached FFT Green's function with one shifted in real space by a user-specified vector. 
- Implemented the `shiftedGreensFunction` method in `FFTOpenPoissonSolver.hpp`, more or less copies/extents `greensFunction` already present in the class. The method is restricted to the HOCKNEY algorithm (and throws an exception if the wrong algorithm is chosen).

### Image charges/Dirichlet example

Added a new integration test, `TestShiftedGreensFunction`, to the CMake test suite, which validates the shifted Green's function implementation by comparing the numerical result (including image correction) against the analytical image-dipole solution for a Gaussian charge near a Dirichlet plane. The test includes detailed diagnostics and passes if both the Dirichlet condition and L2 error are within expected tolerances. 

Note that the "zero plane" is evaluated at the grid point nearest to "0", so the "error" being 4% is exactly as expected. This test only works single rank for now, since flipping a multi-rank field would require a lot more code. 

### Shifted evaluation

An easier test is simply calculating the potential of a bunch located at the origin at a different position. I made such a test (not commited here, since it's just for visualization), where I create a uniform sphere with radius `1e-3` at the origin `(0,0,0)` with `12pC` and calculate the its field with a `1.5e-3` shift in `z` direction. With 64 grid points, the result looks like this:

N64_shifted_greens_z_profile
N64_shifted_greens_yz_sum_overlay

The complete potential is the analytical solution while e.g. the orange line in the 2D plot is the shifted solution of the solver. As one can see, it matches as expected. This should show that at least the shifted Green's function implementation works as expected. The rest of the necessary implementation will either be part of OPALX. Or there will be another PR with a "field flip" helper. 

I will talk about it in detail in today's IPPL meeting.

### DIFF
--- a/src/PoissonSolvers/FFTOpenPoissonSolver.hpp
+++ b/src/PoissonSolvers/FFTOpenPoissonSolver.hpp
@@ -2187,4 +2187,96 @@ namespace ippl {
         }
         ippl::Comm->freeAllBuffers();
     };
+
+    ////////////////////////////////////////////////////////////////////////
+    // Shifted Green's function: fills grn_mr with G(r - shift) on the doubled
+    // grid and caches FFT into grntr_m. Kokkos-parallel, mirrors the HOCKNEY
+    // greensFunction() branch in structure.
+    //
+    // After this call, solve() will convolve rho with the shifted kernel.
+    // To restore the standard kernel, call greensFunction() explicitly.
+
+    template <typename FieldLHS, typename FieldRHS>
+    void FFTOpenPoissonSolver<FieldLHS, FieldRHS>::shiftedGreensFunction(
+        const Vector<double, Dim>& shift) {
+        const int alg = this->params_m.template get<int>("algorithm");
+        if (alg != Algorithm::HOCKNEY) {
+            throw IpplException(
+                "FFTOpenPoissonSolver::shiftedGreensFunction",
+                "Shifted Green's function is only implemented for HOCKNEY.");
+        }
+
+        // Sync mesh spacing with the current RHS mesh (same logic as solve()'s
+        // mesh-change detection). Without this, two failure modes compound:
+        //   1. We would compute the shifted kernel at a STALE hr_m.
+        //   2. A subsequent solve() would see hr_m != mesh->getMeshSpacing(),
+        //      set green=true, and call greensFunction() — overwriting the
+        //      shifted kernel with the standard one.
+        // By updating hr_m (and the dependent mesh2_m / meshComplex_m) here,
+        // solve()'s mesh check finds no change and leaves grntr_m intact.
+        mesh_mp = &(this->rhs_mp->get_mesh());
+        for (unsigned int i = 0; i < Dim; ++i) {
+            hr_m[i] = mesh_mp->getMeshSpacing(i);
+        }
+        mesh2_m->setMeshSpacing(hr_m);
+        meshComplex_m->setMeshSpacing(hr_m);
+
+        const scalar_type pi = Kokkos::numbers::pi_v<scalar_type>;
+
+        typename Field_t::view_type view = grn_mr.getView();
+        const int nghost                 = grn_mr.getNghost();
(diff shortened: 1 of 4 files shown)


aaadelmann (approved):

src/PoissonSolvers/FFTOpenPoissonSolver.hpp
```diff
@@ -2187,4 +2187,96 @@ namespace ippl {
         }
         ippl::Comm->freeAllBuffers();
     };
+
+    ////////////////////////////////////////////////////////////////////////
+    // Shifted Green's function: fills grn_mr with G(r - shift) on the doubled
+    // grid and caches FFT into grntr_m. Kokkos-parallel, mirrors the HOCKNEY
+    // greensFunction() branch in structure.
```
s-mayani: "branch" --> method ?

src/PoissonSolvers/FFTOpenPoissonSolver.hpp
```diff
@@ -2187,4 +2187,96 @@ namespace ippl {
...
+        Kokkos::parallel_for(
+            "Shifted Green's function", grn_mr.getFieldRangePolicy(),
+            KOKKOS_LAMBDA(const int i, const int j, const int k) {
+                // local -> global indices
+                const int ig[3] = {i + ldom[0].first() - nghost,
+                                   j + ldom[1].first() - nghost,
+                                   k + ldom[2].first() - nghost};
+
+                // Half-wrap each axis so FFT cyclic convention gives symmetric
+                // physical offsets around the origin:
```
s-mayani: I am not sure I understand this comment. This is done according to the Hockney method, and has nothing to do with the shift right?

src/PoissonSolvers/FFTOpenPoissonSolver.hpp
```diff
@@ -2187,4 +2187,96 @@ namespace ippl {
...
+
+                // Sign convention matches greensFunction() (HOCKNEY): grn_mr stores -G0
+                // so that solve()'s `rho2tr_m = -rho2tr_m * grntr_m` yields +conv(rho,G0)
+                // where G0(r) = 1/(4 pi |r|).
+                const bool nearSing = (rsq < regThresh);
+                const scalar_type r = Kokkos::sqrt(rsq + nearSing * regThresh);
+                view(i, j, k)       = -1.0 / (4.0 * pi * r);
+            });
+
+        // Fourier-space cache for convolution.
```
s-mayani: Maybe this comment should be more explicit:
"Store Fourier transform of shifted Green's function for convolution in solve()"

s-mayani (changes requested):

src/PoissonSolvers/FFTOpenPoissonSolver.hpp
aliemen: True it's a bit confusing. The half-wrap is Hockney and the per-axis shift is the Green's shift. I'll simplify the comment.

src/PoissonSolvers/FFTOpenPoissonSolver.hpp
aliemen: Done.

src/PoissonSolvers/FFTOpenPoissonSolver.hpp
aliemen: That's indeed better, thanks

s-mayani (approved):

PR #495, opened 2026-04-24
Refactor FFT solver `pack`/`unpack` helpers into a shared header

aliemen:
Move the previously inlined `pack`/`unpack`/`solver_send`/`solver_recv`/` ViewAccess` helpers from `PoissonSolvers`/`FFTOpenPoissonSolver.h(pp)` into a new shared header `Field/FieldBufferOps.hpp` (namespace `ippl::detail`), so other utilities can reuse the same device-parallel, CUDA-aware-MPI packing primitives without duplicating them.

- `FFTOpenPoissonSolver.h`: include the new header, drop the local `ViewAccess` definition.
- `FFTOpenPoissonSolver.hpp`: drop the inlined helpers, pull them into namespace `ippl` via using-declarations so existing call sites remain unqualified.
- Add a value-preserving `pack` variant (`pack_field`/`solver_send_field`) that works for arbitrary element types (e.g. `ippl::Vector`) by copying `view(i,j,k)` directly instead of going through the `Kokkos::complex`/`Kokkos::real` round-trip of the legacy `pack`. 
- `Communicate`/`Tags.h`: add a dedicated `FIELD_MIRROR` MPI tag (`40500`) for clients of the generic field-mirror utility built on top of these helpers. This tag is used from inside OPALX.

I wanted the IPPL field communication helpers accessible from outside (OPALX), so I moved them in their own class. At the moment, I moved the `mirrorField` function into OPALX. You can find it [here](https://github.com/OPALX-project/OPALX/blob/80bb4f5a51c7b7f379103bd4653a3104a771b946/src/PartBunch/FieldMirror.hpp). You can see for example [here](https://github.com/OPALX-project/OPALX/blob/80bb4f5a51c7b7f379103bd4653a3104a771b946/src/PartBunch/FieldMirror.hpp#L118) how the `solver_send`/... operations are reused in OPALX. 

On the IPPL side, this should for now only be a refactor (plus the new communicator tag). Later - once the mirror function is tested more rigorously - I want to move it somewhere as a helper into IPPL.

### DIFF
--- a/src/Field/FieldBufferOps.hpp
+++ b/src/Field/FieldBufferOps.hpp
@@ -58,7 +58,7 @@ namespace ippl {
         };
 
         // Pack a 3D view region into a linear buffer via the real-part of each
-        // cell (legacy FFT-solver semantics: tolerates real or Kokkos::complex
+        // cell (FFT-solver semantics: tolerates real or Kokkos::complex
         // views; discards imaginary parts of complex views).
         template <typename Tb, typename Tf>
         inline void pack(const ippl::NDIndex<3> intersect, Kokkos::View<Tf***>& view,
--- a/src/Field/FieldBufferOps.hpp
+++ b/src/Field/FieldBufferOps.hpp
@@ -1,0 +1,290 @@
+//
+// Shared pack / unpack / solver_send / solver_recv helpers for Field-like
+// communication patterns. Originally inlined in FFTOpenPoissonSolver.hpp and
+// promoted here so other utilities (e.g. mirrorField) can reuse the exact same
+// device-parallel, CUDA-aware-MPI-compatible primitives.
+//
+// All functions operate on Kokkos views and ippl::detail::FieldBufferData
+// buffers whose memory space is inferred from the view, so data stays on the
+// device when the underlying MPI is GPU-aware.
+//
+#ifndef IPPL_FIELD_BUFFER_OPS_HPP
+#define IPPL_FIELD_BUFFER_OPS_HPP
+
+#include <vector>
+
+#include "Types/Vector.h"
+
+#include "Communicate/Archive.h"
+#include "Field/HaloCells.h"
+#include "Index/NDIndex.h"
+
+namespace ippl {
+    namespace detail {
+
+        // Access a view that either contains a scalar, vector, or matrix field
+        // so that the correct element is selected at compile time. Used by
+        // unpack_impl to support rank-0 (scalar), rank-1 (vector), and rank-2
+        // (matrix) field types with a single kernel.
+        template <int tensorRank, typename>
+        struct ViewAccess;
+
+        template <typename View>
+        struct ViewAccess<2, View> {
+            KOKKOS_INLINE_FUNCTION constexpr static auto& get(View&& view, unsigned dim1,
+                                                              unsigned dim2, size_t i, size_t j,
+                                                              size_t k) {
+                return view(i, j, k)[dim1][dim2];
+            }
+        };
+
+        template <typename View>
+        struct ViewAccess<1, View> {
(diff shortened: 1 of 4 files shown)


src/Field/FieldBufferOps.hpp
```diff
@@ -0,0 +1,290 @@
...
+            KOKKOS_INLINE_FUNCTION constexpr static auto& get(View&& view,
+                                                              [[maybe_unused]] unsigned dim1,
+                                                              [[maybe_unused]] unsigned dim2,
+                                                              size_t i, size_t j, size_t k) {
+                return view(i, j, k);
+            }
+        };
+
+        // Pack a 3D view region into a linear buffer via the real-part of each
+        // cell (legacy FFT-solver semantics: tolerates real or Kokkos::complex
```
s-mayani: don't agree with the "legacy" qualifier here

s-mayani (commented):
I agree with all changes, it's a very nice and clean PR. I like the FieldBufferOps.hpp. However, I disagree with the comment on "legacy code". The pack and solver_send with the Kokkos::real stuff still serve a purpose in the original doubling/restricting algorithm, hence, I would not call them legacy.

src/Field/FieldBufferOps.hpp
aliemen: Oh ups, good point. This was from working on OPALX: "legacy semantics" in the sense of not being used anymore for the shifted Greens path in OPALX. But you're right - the wording doesn't make sense in IPPL.

Edit: I simply removed the word "legacy". The "FFT-solver semantics" comment should be valid anyways.

s-mayani:
Also, did you test that the tests still give the right results?
You can run `./TestGaussian_convergence HOCKNEY DOUBLE --info 5` and `./TestGaussian_convergence DCT_VICO DOUBLE --info 5` on multi-rank (I would suggest running 3 ranks and 4 ranks). Reference results are in the reference_outputs directory.

aliemen:
Gaussian `HOCKNEY`:
```
./TestGaussian_convergence HOCKNEY DOUBLE --info 5

> Spacing Error ErrorEx ErrorEy ErrorEz
> 0.25 0.9954200071091134 0.9986417059820205 0.9986417059820202 0.9986417059820202
> 0.125 0.3013269357499338 0.7508364127190535 0.7508364127190549 0.7508364127190545
> 0.0625 0.052811160790769 0.3024289005995584 0.3024289005995603 0.3024289005995602
> 0.03125 0.01378314255797071 0.08052796496843252 0.08052796496843305 0.08052796496843317
> 0.015625 0.003489811209812147 0.02048590401578897 0.02048590401578951 0.02048590401578953
> 0.0078125 0.0008764090082055583 0.005150608932780393 0.005150608932780661 0.005150608932780617
```
and `DCT_VICO`:
```
./TestGaussian_convergence DCT_VICO DOUBLE --info 5

> Spacing Error ErrorEx ErrorEy ErrorEz
> 0.25 0.9948059296399622 0.996767163711261 0.9967671637112608 0.996767163711261
> 0.125 0.2433765361633136 0.4549151699608537 0.454915169960854 0.454915169960854
> 0.0625 0.0002889522144481186 0.01861181163792976 0.01861181163792988 0.01861181163792984
> 0.03125 1.174356939339496e-09 5.811484600113261e-07 5.811484600069108e-07 5.811484600056847e-07
> 0.015625 1.794041904881624e-15 3.716910027294508e-15 3.926486592539179e-15 3.937196482628669e-15
> 0.0078125 2.717993979626873e-15 7.516436685741153e-15 7.468010494813765e-15 7.488192737045828e-15
```

This was single rank. 3 and 4 ranks locally on my mac (openmp) gives the same results up to machine precision. From what I saw, it also matches the reference solution well.

s-mayani:
Perfect, will approve.

s-mayani (approved):

PR #499, opened 2026-04-29
Fix Doxygen warnings in IPPL API comments

aaadelmann:
Clean up a first batch of API documentation issues reported by Doxygen:
- fix over-qualified Communicator::split definition
- remove duplicate FDTD constructor docs from implementation files
- correct stale and mismatched @param names
- replace malformed @brieg command with @brief
- avoid recursive or incorrect copydoc targets in buffer docs
- add missing parameter names for particle, FEM, and solver APIs

Regenerating Doxygen now removes the targeted warning families and keeps the manual/API link check clean.


AGENTS.md
```diff
@@ -0,0 +1,22 @@
+# AGENTS.md
+
+## Repository purpose
+IPPL is a C++ Particle and fields Framwork. Preserve physical correctness over stylistic cleanup.
```
biddisco:
```suggestion
IPPL is a C++ Particle and fields Framework. Preserve physical correctness over stylistic cleanup.
```

biddisco (commented):
LGTM apart from one typo. I don't know why the clang-format relative path fix is needed -was it broken somewhere? (I looked at the new version but couldn't quite see what it is doing, however, I trust you wouldn't put it there without checking it ...)

biddisco (approved):
LGTM

PR #501, opened 2026-05-04
PIF, Higher Order and Optimized Scatter/Gather, 

PaulFisch:
This PR adds the Particle-in-Fourier method to IPPL, together with restructured FFT, scatter/gather, particle layout, communication, and timing infrastructure. A list of changes:

### Build system

- New `IPPL_ENABLE_FINUFFT` (CPU finufft / GPU cufinufft via FetchContent)
  and `IPPL_ENABLE_CUFFTMP` (cuFFTMp + NVSHMEM) build flags, both default OFF.
- Kokkos default bumped to 5.0.0
- `cmake/AddIpplTest.cmake` passes `TEST_LINK_LIBS` through to non-integration
  tests.
- `cmake/AutoTunePresets.cmake`, `IpplAutoTunePresets.h.in`, and pre-tuned
  sweeps under `cmake/auto_tune/sm_90/` ship the autotuner data alongside
  the build (for new scatter/gather kernels).

### FFT refactor

- The old `FFT.h/.hpp` is split into a Backend layer
  (`Backend/{Backend,CuFFT,CuFFTMp,Heffte}.h`) handling plan
  management, and a Transform layer
  (`Transform/{CC,RC,PrunedCC,PrunedRC,Trig,NUFFT}.h`) exposing per-transform
  front-ends.
- A new native NUFFT (Type-1 and Type-2) lives under `src/FFT/NUFFT/`:
  - `NativeNUFFT.h` drives the upsampling-FFT pipeline.
  - `ESKernel.h` is an exponential-of-semicircle spreading kernel with
    width-specialised optimized evaluation.
  - `Correction.h` provides deconvolution factors, `NUFFTUtilities.h`
    provides helpers.
- The new NUFFT type plugs into the Transform layer via `NUFFT.{h,hpp}`
  and supports both the native engine and the (cu)FINUFFT backend behind
  `IPPL_ENABLE_FINUFFT`.

### Interpolation

- New scatter dispatch in `Interpolation/Scatter/`: `Scatter.h` selects
  between `AtomicScatter`, `GridParallelScatter` (output-focused), and
  `TiledScatter` back-ends through `ScatterConfig` and a runtime
  `TileSizeCache` loaded from the autotune presets.
- Matching gather front-end (`AtomicGather`, `AtomicSort` variants) under
  `Interpolation/Gather/`.
- `Binning.h` bin-sorts particles by tile so hot tiles avoid atomic
  contention. `Kernels.h`, `WidthDispatcher.h`, and `CoordinateTransform.h`
  are shared with the native NUFFT.
- `AutoTune.{h,cpp}` and `TileSizeCache.cpp` provide the runtime
  cache-lookup that lets the same binary pick a different tile size on
  different hardware.

### Particle layout

- `ParticleSpatialLayout::update` is rewritten from the per-iteration
  vector-of-vectors model into a packed locate + single-pass send/recv
  pipeline with reusable scratch (`rankSendCount`, `sendOffsets`, `sendIds`,
  `cursor`). Buffers are reused across updates and host-side hashes are
  avoided.
- `ParticleSort.h` / `SortBuffer.h` provide a bin-sort with a shared
  device-side buffer pool (`BinSortBuffers`) so repeated sorts in
  scatter/gather hot paths reuse storage.
- `FieldLayout` now stores `nghost`; `HaloCells.{h,hpp}` thread it through
  `accumulateHalo` / `fillHalo`, enabling kernel-width-aware halos (>1
  ghost cell). (Before this PR, nghosts>1 was broken in IPPL)
- `ParticleAttrib` exposes `scatterPIFNUFFT` / `gatherPIFNUFFT` guarded by
  `IPPL_ENABLE_FFT`.
- `ParticleAttrib::scatter(...)` / `ParticleAttrib::gather(...)` (CIC overloads used by 
  every alpine PIC example) now
  forward to the kernel-aware path with `LinearKernel<MeshType>` and a
  default `ScatterConfig` / `GatherConfig`. PIC paths therefore inherit the
  same atomic / atomic-sort / tiled / output-focused dispatch and
  `TileSizeCache` lookup as NUFFT.

### Communication and Particle Update

This PR removes a per-attribute `buf_m` send/recv staging view for every
`ParticleAttrib`. The old update path went
`attribute view -> per-attribute buf_m -> MPI buffer → per-attribute buf_m->→
attribute view`, with one `buf_m` per attribute per exchange and copies on
both ends. The new path serializes attribute data directly into a shared,
reused MPI buffer managed by a singleton `BufferHandler`, and
deserializes straight back into the destination attribute's storage at a
caller-supplied offset. The per-attribute `buf_m` is not needed for the common path.

- `ParticleAttrib::serialize` / `deserialize` now operate on `dview_m`
  directly and accept a `hash` view (gather-on-the-fly of the live
  particle indices going to one rank) and a destination `offset`. There is
  no intermediate per-attribute buffer to size, allocate, copy into, or
  copy out of. `getView()` returns a subview over the live range only, so
  attribute kernels never see the overallocated tail.
- A granularity-aware GPU allocator is added to `Archive`'. Device
  archives are allocated with raw `cudaMalloc` / `hipMalloc` so the
  pointer is page-aligned. Host
  archives keep using a regular `Kokkos::View`.

### Utilities

- `IpplTimings.{h,cpp}`: track per-call measurements alongside cumulative
  wall time, support `resetAllTimers()` (for warmup), and add
  `dumpToCSV()` for offline analysis. `print()` reports measurement counts
  and per-occurrence statistics (mean, stddev, min, max).
- `BufferView.h`: lightweight, type-erased view over a contiguous device
  byte buffer, used by the new send/recv path.
- `Tuning.h`: `TileSizeTuner`  interface to Kokkos Tools autotuner that
  enumerates tile-size candidates for the new scatter/gather back-ends.
- `ParallelDispatch.h`: helpers for selecting parallel-for vs.
  parallel-reduce policies based on view rank/extents and execution space.

### Alpine

- `alpine/ElectrostaticPIF/{LandauDampingPIF, LandauDampingPIFPruned,
  BumponTailInstabilityPIF, PenningTrapPIF}.cpp`: added as Alpine PIF examples.
- `alpine/ElectrostaticPIF/{ChargedParticlesPIF, ChargedParticlesPIFPruned}.hpp`:
  particle-bunch container specialised for the PIF schemes

### Unit tests

- `FFT/NUFFT.cpp` + `NUFFTTestUtils.h`: test matrix for NUFFT Type-1
  and Type-2 (small / medium grid, with / without upsampling, three spread
  methods, tolerance sweep at 1e-4 / 1e-7 / 1e-10, FINUFFT cross-check if compiled,
  three gather methods).
- `FFT/NUFFTAccuracy.cpp`: tolerance-vs-error sweep that compares the
  native NUFFT against a direct DFT reference.
- `Interpolation/KernelGatherScatterTest.cpp`: charge conservation,
  periodicity, gather-of-constant-field, scatter/sort comparison,
  roundtrip; covers Atomic, Tiled, OutputFocused
- `Interpolation/Binning.cpp`: bin-sort correctness and stability tests.
- `Particle/ParticleUpdate.cpp`: spatial-layout update tests under uniform
  decomposition.
- `Particle/ParticleUpdateNonuniform.cpp`: same for non-uniform domain
  decompositions

### Integration tests

- `test/FFT/TestFFTCC.cpp` extended for the new pruned variants and the
  heFFTe backend trait.
- `test/particle/benchmarkParticleUpdate.cpp` switched to
  `Kokkos::Random_XorShift64_Pool` device init (avoids host->device copies
  between iterations) and adapted to the new `ParticleSpatialLayout`
  signature.


alpine/ParticleContainer.hpp
```diff
@@ -32,9 +32,9 @@ class ParticleContainer : public ippl::ParticleBase<ippl::ParticleSpatialLayout<
 
     void registerAttributes() {
         //only needed for vis
-        P.set_name("velocity");
-        q.set_name("charge");
-        E.set_name("electric_field");
+        // P.set_name("velocity");
+        // q.set_name("charge");
+        // E.set_name("electric_field");
```
srikrrish: Why are these commented? Is this feature removed?

cmake/Dependencies.cmake
```diff
@@ -317,3 +320,91 @@ if(IPPL_ENABLE_TESTS)
...
+    add_compile_definitions(ENABLE_GPU_NUFFT)
+    add_compile_definitions(FINUFFT_USE_CUDA)
+  endif()
+  set(FINUFFT_USE_CPU ON CACHE BOOL "")
+
+  # cufinufft's CUDA RDC fatbin registration segfaults at startup when its
+  # device code is wrapped in a .so. Force static for the FetchContent build
+  # regardless of the global BUILD_SHARED_LIBS setting.
+  set(_ippl_saved_bsl ${BUILD_SHARED_LIBS})
+  set(BUILD_SHARED_LIBS OFF)
```
srikrrish, line 347: Does this overcome the problem of `FINUFFT_STATIC_LINKING=off` that we needed to set before?

src/FFT/Backend/Heffte.h
```diff
@@ -0,0 +1,247 @@
...
+            using complex_t   = Kokkos::complex<T>;
+            using backend_t   = typename HeffteBackend<MemSpace>::c2c;
+            using heffte_t    = heffte::fft3d<backend_t, long long>;
+            using workspace_t = typename heffte_t::template buffer_container<complex_t>;
+
+            // Standard constructor
+            HeffteC2C(const heffte::box3d<long long>& inbox, const heffte::box3d<long long>& outbox,
+                      MPI_Comm comm, const ParameterList& params, int maxBatchSize = 1)
+                : maxBatchSize_(maxBatchSize) {
+                static_assert(Dim == 3, "HeFFTe wrapper only supports 3D");
```
srikrrish: Does this mean we cannot do 2D (I know Heffte doesn't work with 1D though)? My memory is we specify box size as 1 to perform 2D FFT in the previous code. Isn't it? Also I see that R2C doesn't have this assert.

src/FFT/NUFFT/ESKernel.h
```diff
@@ -0,0 +1,446 @@
...
+    namespace nufft {
+
+        // ============================================================
+        // ES Kernel polynomial evaluation for each width
+        // Coefficients are inlined to work in device code
+        // ============================================================
+
+        // w = 4, beta = 9.2, degree = 5, error < 3.53e-4
+        template <typename T>
+        KOKKOS_INLINE_FUNCTION T es_kernel_eval_w4(T x) {
```
srikrrish, line 39: Why there are no kernels with 1 <= w < 4 ?

unit_tests/FFT/NUFFT.cpp
```diff
@@ -0,0 +1,781 @@
+//
+// Unit test NUFFT
+//   Type-1 (particles -> grid) and Type-2 (grid -> particles).
+//   Parameterised over precision (double, float) and dimension (2, 3).
+//
+//   Limitations:
+//   - FINUFFT-backed cases are 3D-only because the FINUFFT FFT field type
+//     is hard-coded as 3D in src/FFT/Transform/NUFFT.h.
+//   - Native NUFFT internals (Correction.h, NativeNUFFT.h spread/gather
+//     plumbing) are also currently hard-coded for Dim == 3, so 2D test
```
srikrrish: This means if we want to test for 2D then it has to be changed manually in these locations?

srikrrish (changes requested):

srikrrish:

alpine/ParticleContainer.hpp
PaulFisch: I will add back. I commented out temporarily when I originally forked, because there were std::strings in the name attribute, and that causes compilation issues. That was solved in between.

cmake/Dependencies.cmake
PaulFisch: It seems (I hope) yes, at least on my laptop. I am not 100% sure if it works all the time..

src/FFT/Backend/Heffte.h
PaulFisch: This should be removed and reverted to the previous 2D/3D. Before also, there was no other than 2D and 3D in heffte. I tested the gather/scatter, but forgot this - will add back.

src/FFT/NUFFT/ESKernel.h
PaulFisch: For w<4 (I.e. tolerance 1e-3 and lower), it does not really matter - we anyways have so low arithmetic intensity that we are limited by cache.

unit_tests/FFT/NUFFT.cpp
PaulFisch: I will patch back the 2D, I changed at some point when writing the paper and forgot when I assembled the PR.

PaulFisch:
I still have a commit fixing some of the compile failures on juwels, I will also address your comments @srikrrish

PaulFisch:
Here is some scaling of LandauDamping PIC (512^3 on Alps, density 8) of this PR vs current master 
scaling

aaadelmann:
Update looks great. What are the secrets? Is the update special to the PIF?

PaulFisch:
> Update looks great. What are the secrets? Is the update special to the PIF?

No, that is PIC scaling, not PIF, I clarified above. Mainly I removed the temporary allocations that happen in every update iteration, I also restructured the update algorithm a bit.

I don't have permissions to start the CSCS CI, someone else needs to do.

The gather/scatter don't really see much change because most of my work was on the higher-order kernels and optimizing them, especially for strong scaling where the local sizes are relatively small at larger node counts, there is not much too optimize in scatter and gather at PIC (i.e. with width one). At higher order kernels (for the NUFFT or potentially higher order PIC made possible with this PR), all the stuff I added for Scatter/Gather matters significantly more.

aaadelmann:

aaadelmann:
CSCS CI is running. I will do also a code rundown.

aaadelmann (changes requested):
Doxygen coverage is inconsistent. Several important new public implementation files have essentially no Doxygen: AutoTune.cpp, GridParallelScatter.h, AtomicScatter.h, TiledScatter.h, CuFFT.h, CuFFTMp.h, PrunedCC.h, PrunedRC.h, ParticleSpatialLayout.hpp, Archive.hpp, and the Alpine PIF classes.

PaulFisch:
> Doxygen coverage is inconsistent. Several important new public implementation files have essentially no Doxygen: AutoTune.cpp, GridParallelScatter.h, AtomicScatter.h, TiledScatter.h, CuFFT.h, CuFFTMp.h, PrunedCC.h, PrunedRC.h, ParticleSpatialLayout.hpp, Archive.hpp, and the Alpine PIF classes.

True I should work on some more documentation. Also, I should probably add at least some documentation to the new IPPL Manual, I will try to do by tomorrow afternoon.

alpine/ElectrostaticPIF/ChargedParticlesPIF.hpp
```diff
@@ -0,0 +1,610 @@
+// ChargedParticlesPIF header file
```
aaadelmann, line 1: git diff --check reports trailing whitespace in ChargedParticlesPIF.hpp (line 535). There are also long lines and local comments like “Could definiitely be improved” in Scatter.h (line 121), which should not go into production code.

aaadelmann (changes requested):
Complexity is super high for one PR. The PR combines FFT backend refactor, NUFFT, scatter/gather redesign, particle migration rewrite, autotuning, timing changes, build-system changes, examples, and many tests. That makes review and regression attribution difficult. 

Would it be possible to have the FFT as a separate PR?

PaulFisch:
> Complexity is super high for one PR. The PR combines FFT backend refactor, NUFFT, scatter/gather redesign, particle migration rewrite, autotuning, timing changes, build-system changes, examples, and many tests. That makes review and regression attribution difficult.
> 
> Would it be possible to have the FFT as a separate PR?

Thanks for the feedback, and I do see the point about the PR being large. Before going through the individual pieces, some context that I hope is useful: I have put a considerable amount of time into preparing this PR, unpaid and without other compensation, upon specific and repeated request from your team. I was very glad to do so, but I do also have a job and a master thesis running in parallel. If there is no interest in merging the changes after all, I would really have appreciated hearing that sooner, before I spent tens of hours on restructuring.

Going through the components:

- The FFT (where I assume the NUFFT is meant) relies quite directly on the Scatter/Gather redesign, as higher-order Scatter/Gather is at the core of a NUFFT. Without it, there is no NUFFT. The algorithm for width-1 (i.e. PIC) does not change. Since IPPL previously had no support for higher-order Scatter/Gather, it is unfortunately not possible to merge the NUFFT without also adding the Scatter/Gather work.
- The FFT refactor itself can easily be undone by merging all the currently included headers in `FFT.h` back into a single ~2000-LOC file, since I did not significantly change the FFT code itself but only broke it up into pieces. Happy to do that if it would help review.
- The particleUpdate is indeed separate and I can pull it out of the PR relatively easily. As shown in the plot above, it gives a substantial performance boost, and I also added a significant amount of tests to the particle update which surfaced a few minor bugs in the previous version. Happy to revert if that is preferred.
- The Auto-Tuning feature is used solely for the higher-order Scatter/Gather kernels, which do not currently exist in IPPL. It could be removed, at the cost of noticeably decreased performance. It selects optimal scatter/gather strategies based on local particle density, and was something Sri was particularly interested in for imbalanced distributions.
- I can remove the timing changes if desired. I would still mention that timings generated without warmup, which was the case before, can be quite misleading. For PIC, a significant share is spent in IPC buffer registration on Alps.
- The build system changes are quite limited: only what is needed for the higher-order Scatter/Gather kernels, a Kokkos default bump to 5.0.0 (which I can revert), and the optional cuFINUFFT dependency. I can also remove the cuFINUFFT path entirely, though it does perform better than the alternative in some cases.
- I can certainly remove tests from the PR, though I would push back on that one. Especially for a PR of this size, the test coverage should be kept.
- Regarding the "this could be better" comments in `Scatter.h`, I am happy to remove them if preferred. To clarify what they are about: they refer to strategies for *selecting* between scatter algorithms, not to the algorithms themselves. Every algorithm in there is faster than what was in IPPL before. The comments are simply meant as hints for future contributors who might find even better selection heuristics, since I am sure better ones exist. If the view is that the code is not fast enough for IPPL without that further improvement, I would just note that by the same standard, the previous scatter/gather and particle update paths would also not have qualified. Either way, happy to take the comments out.

aaadelmann:
I totally  understand your boundary conditions. No one has said that we will not merge or that you should undo anything - but again it is simply a complexity problem.  Complexity in reviewing and testing. Consider my comment on breaking up as "resolved".  If we really need this we can do the split up the merge my ourself.

srikrrish:
Currently the Jülich's CI/CD have some token issues (in the final results pushing to the web page) and me and my colleague are looking into it. But I was able to see the results of the tests directly in gitlab and they were all passing. I now see that CSCS's CI is also passing. @PaulFisch We greatly appreciate your work and time on this and in case you find a bit of time to improve the documentation that would be great. But otherwise also we can understand. 

@aliemen Could you may be pull this branch and check the open solver or whatever is relevant for you/OPALX? @aaadelmann Could you may be run some tests on LUMI as you mentioned before? 

Since Paul has already posted results on Alps, we will be then good to merge in my opinion.

PaulFisch:
I can perform some LUMI scalings, I already have the binaries compiled there. It would probably indeed be good to check some applications before merging this.

srikrrish:
> I can perform some LUMI scalings, I already have the binaries compiled there. It would probably indeed be good to check some applications before merging this.

That would be great. Thanks @PaulFisch . As I mentioned before @aliemen and/or @aaadelmann can check some OPALX relevant stuffs I think.

aliemen:
@srikrrish yes, I will compile OPALX against the branch and try to run our regression tests. Will come back with the results probably later today or tomorrow morning.

aaadelmann:
Sonali will run all her Solver tests also on high node counts

PaulFisch:
I also just found a bug in both my branch and the master with the overallocation being cast to integers everywhere. I have used in the alps PIC run a non-integer overallocation for both master and my branch. I will push that together with the documentation and add an updated scaling here when the runs finish.

On LUMI, both master and my branch segfault after a few iterations. I am unsure if this is due to FS issues today (https://lumi-supercomputer.eu/lumi-service-status/information-lustre-filesystem-performance/), the heffte issue, OOM, or something else. I am trying to triage.

Happy to help/discuss in case there is regressions anywhere and we need to split up the PR in some way. I had conflicting meetings on Tuesdays the last weeks unfortnately, but also happy to set up a meeting if required.

cmake/AutoTunePresets.cmake
```diff
@@ -0,0 +1,140 @@
...
+    endif()
+  endif()
+
+  # Bake into a generated header consumed by TileSizeCache / GatherCache.
+  set(IPPL_AUTOTUNE_PRESET_DIR "${_dst_dir}")
+  set(IPPL_AUTOTUNE_ARCH_TAG "${_tag}")
+  configure_file(
+    "${CMAKE_SOURCE_DIR}/cmake/IpplAutoTunePresets.h.in"
+    "${CMAKE_BINARY_DIR}/include/IpplAutoTunePresets.h"
+    @ONLY)
```
aliemen, line 138: This breaks when IPPL is consumed via `FetchContent` from OPALX because `CMAKE_SOURCE_DIR` points to the top-level OPALX source directory, not the IPPL source directory. In that case `AutoTunePresets.cmake` tries to read:

```cmake
${CMAKE_SOURCE_DIR}/cmake/IpplAutoTunePresets.h.in
```

which resolves to `opalx/cmake/IpplAutoTunePresets.h.in`, but the template actually lives in IPPL's own `cmake/` directory.

A fix is to resolve paths relative to `AutoTunePresets.cmake` itself, e.g.

```cmake
set(_ippl_cmake_dir "${CMAKE_CURRENT_FUNCTION_LIST_DIR}")
set(_src_dir "${_ippl_cmake_dir}/auto_tune/${_tag}")

configure_file(
  "${_ippl_cmake_dir}/IpplAutoTunePresets.h.in"
  "${CMAKE_BINARY_DIR}/include/IpplAutoTunePresets.h"
  @ONLY)
...
```
With this suggestion, the cmake runs fine. Still doesn't compile - but that's the Kokkos 5 issue. Will comment again when I succeeded.

PaulFisch: I will add that. Is OpalX incompatible with Kokkos 5.x, or is there something specific that I do with it? I think I am not using any kokkos 5.x features, I should be able to bump down the version easily

PaulFisch: That is, I was bumping the version because Kokkos 5.x improves performance on some relevant parts (in particular atomics), so long-term it is probably nice to have Kokkos 5.x in OPALX, but it should not block merging this, I can just bump down the version again

src/Interpolation/Gather/AtomicGather.h
```diff
@@ -0,0 +1,113 @@
...
+        static constexpr bool requires_binning = UseSorting;
+        static constexpr unsigned Dim          = Types::Dim;
+
+        using RealType        = typename Types::RealType;
+        using ValueType       = typename Types::ValueType;
+        using memory_space    = typename Types::memory_space;
+        using execution_space = typename Types::execution_space;
+
+        struct Arguments : GatherArgumentsBase<Arguments, Types> {
+            using PermuteView = Kokkos::View<uint64_t*, memory_space>;
```
aliemen: In some places, `Kokkos::View<uint64_t*, memory_space>` and `Kokkos::View<ippl::detail::size_type*, memory_space>` are mixed (maybe that's just a problem, because OPALX uses `size_type` more often...). I think it would be nice to use `size_type` everywhere.

On macOS this fails because `std::size_t` is `unsigned long`, while `uint64_t` is `unsigned long long`, so these are different Kokkos view value types and cannot be assigned to each other.

PaulFisch: I will try to address. The problem is that size_type is not size_t, but on some exec spaces a 32-bit integer, and the particle number can easily exceed maxint.

cmake/AutoTunePresets.cmake
aliemen: You're right, I thought most of the other errors where Kokko-5, but it works with Kokkos 4. With my fixes below (using `size_type` everywhere) I was able to compile it. OPALX still crashed, but I am looking into it.

src/Interpolation/Gather/AtomicGather.h
PaulFisch: Ah, but the uint64_t is definitely a bug, it should be size_t everywhere for wider int.

PaulFisch: Actually, do you mean ippl::size_type or the Kokkos size_type that is a typedef of the exeuction spaces? I can try to use uniform ippl::size_type, that is surely the cleanest solution, that is also probably what you meant.

src/Particle/ParticleAttrib.h
```diff
@@ -112,11 +135,22 @@ namespace ippl {
...
-        const view_type& getView() const { return dview_m; }
+        /*!
+         * Returns a subview of the underlying storage covering only the live
+         * particle range [0, size()).
+         */
+        view_type getView() {
+            return Kokkos::subview(dview_m,
+                                   Kokkos::make_pair(size_type(0),
+                                                     static_cast<size_type>(*(this->localNum_mp))));
+        }
```
aliemen, line 146: Alright, this change was also a problem: we use something like "`container->R.getView().extent(0)`" to check the overall allocated size (used to optimize `alloc` vs `create` calls). This doesn't work when the `extent` is limited to the subview leading to unexpected behaviour...

However, I saw that ".size()" is available from the outside. Will patch this in OPALX instead (using `size` might be cleaner semantics anyways).

aliemen:
Alright, after the `size_type` fix, the `cmake` path fix and the `R.extent(0)` vs `R.size()` fix, I was able to compile and run OPALX. Then, I had to fix a unit test that couldn't handle the temporary subview returned by `getView` - but that was an OPALX problem again. 

Now, all of our unit and regression tests passed cleanly on CPU and GPU (GH200 on Daint). Our regression tests only test single rank. But from my local runs, all of them looked good with up to 8 ranks (CPU).

_Note that I pushed a branch here to IPPL, `test-pr-501`, that I used to compile OPALX against where I applied my fixes. Feel free to look at it if you want to see what would work with OPALX. Changes I made to OPALX are in [`412-...`](https://github.com/OPALX-project/OPALX/tree/412-opalx-does-not-yet-work-with-kokkos-5), will also make PR as soon as possible._

s-mayani:
Alps scaling studies, using the same configurations (512^3 and 1024^3) as the paper:

[TOTAL_scalings_1024_paper_alps.pdf](https://github.com/user-attachments/files/27512377/TOTAL_scalings_1024_paper_alps.pdf)
[TOTAL_scalings_512_paper_alps.pdf](https://github.com/user-attachments/files/27512378/TOTAL_scalings_512_paper_alps.pdf)


I have added a line comparing with the paper. The FFT clearly benefits from the update's improvement. Something that caught my eye is in the 1024^3 case, at the very end (512 GPUs) this branch ("Paul" in the legend) seems to flatten and stop scaling, whereas in the paper this was not the case. I don't know why this is.

srikrrish:
> Alps scaling studies, using the same configurations (512^3 and 1024^3) as the paper:
> 
> [TOTAL_scalings_1024_paper_alps.pdf](https://github.com/user-attachments/files/27512377/TOTAL_scalings_1024_paper_alps.pdf) [TOTAL_scalings_512_paper_alps.pdf](https://github.com/user-attachments/files/27512378/TOTAL_scalings_512_paper_alps.pdf)
> 
> I have added a line comparing with the paper. The FFT clearly benefits from the update's improvement. Something that caught my eye is in the 1024^3 case, at the very end (512 GPUs) this branch ("Paul" in the legend) seems to flatten and stop scaling, whereas in the paper this was not the case. I don't know why this is.

Thanks a lot Sonali. Yeah the FFT PIC is vastly improved now! For the flattening at 512 GPUs (for 1024^3 case) do you mean the PCG solver? Because I don't see that in FFT or FEM.

s-mayani:
> Thanks a lot Sonali. Yeah the FFT PIC is vastly improved now! For the flattening at 512 GPUs (for 1024^3 case) do you mean the PCG solver? Because I don't see that in FFT or FEM.

Yeah sorry I mean only the PCG solver.

srikrrish:
> > Thanks a lot Sonali. Yeah the FFT PIC is vastly improved now! For the flattening at 512 GPUs (for 1024^3 case) do you mean the PCG solver? Because I don't see that in FFT or FEM.
> 
> Yeah sorry I mean only the PCG solver.

Could you may be just submit the run corresponding to that data point only to see if it may be some one time effect?

PaulFisch:
Generally the PCG performance decreased, right? Can you send me the script that you submitted with, then I can try to analyze the cause

After hours of debugging yesterday night, I found that heffte is fundamentally broken with the ROCM/MPICH combination that we pass as linker flags in the LUMI IPPL builds. I was able to reproduce in a standalone project without Kokkos. The segfault has nothing to do with my branch and just seems to appear more often for some reason in my branch than in master. When enabling enough HSA debugging env variables, the segfault happens reliable in the first warmup FFT. I will try building on different environments to see if that makes it work. Unsure whether this is a heffte, Cray MPICH, HSA or rocm compiler bug.

srikrrish:
> Generally the PCG performance decreased, right? Can you send me the script that you submitted with, then I can try to analyze the cause
> 
> After hours of debugging yesterday night, I found that heffte is fundamentally broken with the ROCM/MPICH combination that we pass as linker flags in the LUMI IPPL builds. I was able to reproduce in a standalone project without Kokkos. The segfault has nothing to do with my branch and just seems to appear more often for some reason in my branch than in master. When enabling enough HSA debugging env variables, the segfault happens reliable in the first warmup FFT. I will try building on different environments to see if that makes it work. Unsure whether this is a heffte, Cray MPICH, HSA or rocm compiler bug.

Thanks a lot. Seems something that @paubauer may be able to help. The standalone reproducer could be quite useful for him. No need to spend a lot of time on this especially if it seems to lead to a rabbit hole.

s-mayani:
I launched the job for that specific data point, it's in the queue. As for the heffte issue, seems related to Issue #481 .

src/Interpolation/Gather/AtomicGather.h
aliemen: Ah yes, I was talking about `ippl`'s `size_type`. In OPALX we aliased something like `using size_type = ippl::detail::size_type`, so I was talking about that one.

src/Interpolation/Scatter/AtomicScatter.h
```diff
@@ -0,0 +1,364 @@
...
+
+            // helper: fill weights for a compile-time dimension D using g0 computed on-demand
+            auto fill_dim = [&](auto Dtag, RealType g0d) {
+                constexpr int D = decltype(Dtag)::value;
+                Kokkos::parallel_for(Kokkos::ThreadVectorRange(team, W), [&](const int i) {
+                    const RealType xi = g0d - RealType(i) * inv_hw;
+                    if constexpr (Types::KernelType::has_width_template)
+                        kw_ptr[D * W + i] = args.kernel.template eval<W>(xi);
+                    else
+                        kw_ptr[D * W + i] = args.kernel(xi);
```
aliemen, line 235: For some reason, this is a problem with cuda 12.1 and gcc 12 on Gwendolen. I cannot compile OPALX against it:

```
.../cuda-ampere80/build/_deps/ippl-src/src/Interpolation/Scatter/AtomicScatter.h(218): error: class "ippl::Interpolation::LinearKernel<double>" has no member "eval"
                          kw_ptr[D * W + i] = args.kernel.template eval<W>(xi);
                                                                   ^
```
CUDA 12.1 with GCC 12 is diagnosing the discarded `eval<W>` expression inside a Kokkos device lambda/team kernel. Newer CUDA, like 12.9 with GCC 14, apparently handles that template/device-lambda case correctly... 

We could either find a workaround (_like making all interpolation kernels provide a harmless width-templated wrapper_) or we say "doesn't work with cuda/12.1" and ignore it. But that would mean we need a new module on Gwendolen (which is the case for [Kokkos 5 anyways](https://github.com/OPALX-project/OPALX/pull/414#issuecomment-4406060302)). @aaadelmann what's your opinion?

PaulFisch: There were still a lot of compiler bugs with this kind of expressions in cuda 12.9, probably it's not too difficult to build a workaround on my side, I will check.

s-mayani:
The new run still produced a similar timing, with the PCG flattening at 512 GPUs.

aaadelmann:
@s-mayani if you still have time until departure, send me all timing files.

PaulFisch:
I tried to look a but at PCG profiles - most time is spent in the cudaMallocs and cudaFree in every iteration. On my tests, my branch was a bit faster, but maybe I used a different preconditioner. I have two theories: a) My branch does allocation of other stuff differently, and this affects the timing of allocations in the PCG, b) My branch needs more iterations to converge for some reason. I will investigate further. Likely the best long-term solution would be to pre-allocate the views in the PCG

srikrrish:
> I tried to look a but at PCG profiles - most time is spent in the cudaMallocs and cudaFree in every iteration. On my tests, my branch was a bit faster, but maybe I used a different preconditioner. I have two theories: a) My branch does allocation of other stuff differently, and this affects the timing of allocations in the PCG, b) My branch needs more iterations to converge for some reason. I will investigate further. Likely the best long-term solution would be to pre-allocate the views in the PCG

More iterations to converge seems to be the most concerning thing for me. @s-mayani Do you know whether you are also seeing the same behaviour in your tests?

PaulFisch:
I should clarify that I did not actually see my branch needing more iterations to converge. I was just speculating what might be the issue for the difference in performance. If it is the mallocs, the cleaner way is likely to slightly refactor the PCG not to malloc in every iteration of the CG, as that would give a massive performance boost. Given that I did not observe a slow down within an iteration in my tests, I just stated that theoretically it could be that my branch needs more iterations, which would also explain longer runtime. That would be something that I would need to investigate, if that would be the case.

aaadelmann:
If you could paste the timing-profiles into this PR, I will make an issue out of it.

PaulFisch:
I have now fixed all the reviews and tried to address the PCG problem. 
scaling
I measured now a speedup on PCG, let me know if there is still a regression observed. I just realised that I did not update the branch wrt to the master, will try to merge with the master.

srikrrish:
> I have now fixed all the reviews and tried to address the PCG problem. scaling I measured now a speedup on PCG, let me know if there is still a regression observed. I just realised that I did not update the branch wrt to the master, will try to merge with the master.

This looks great! Thanks a lot @PaulFisch !

PaulFisch:
Will see if CI passes now, otherwise will try to fix. The two problems discussed earlier with hangs and PCG slowdown seemed to be connected - I was not able to reproduce the deadlock without PCG anymore, and with my updates to PCG, it did not hang anymore. I now modified so that PCG does not do any rellocations within the CG loop, but I pushed that (i.e. asked an LLM to squash onto) a single commit, so it should be somewhat easily separable from the rest.

aaadelmann:
Awesome !!!!!

srikrrish:

aaadelmann:
@PaulFisch are you done with working on this PR?

PaulFisch:
Yes, I think it should work now, I saw the mi300 CI failed due to some CI issues - let me know if there are any problems

aaadelmann:
This gigantic PR is divided and committed into 

002b814aa Merge pull request #572 from IPPL-framework/pr501-pif-final
758eca511 Merge pull request #571 from IPPL-framework/pr501-nufft-final
0424793e7 Merge pull request #570 from IPPL-framework/pr501-fft-final
1b66ed9e9 Merge pull request #569 from IPPL-framework/pr501-hosg-final
73efeee5b Merge pull request #544 from IPPL-framework/pr501-communication-particle-update
b35146d51 Merge pull request #543 from IPPL-framework/pr501-pcg

aaadelmann:
This PR was disected into 

002b814aa Merge pull request #572 from IPPL-framework/pr501-pif-final
758eca511 Merge pull request #571 from IPPL-framework/pr501-nufft-final
0424793e7 Merge pull request #570 from IPPL-framework/pr501-fft-final
1b66ed9e9 Merge pull request #569 from IPPL-framework/pr501-hosg-final
73efeee5b Merge pull request #544 from IPPL-framework/pr501-communication-particle-update
b35146d51 Merge pull request #543 from IPPL-framework/pr501-pcg

and then merged.

aaadelmann:
This PR was disected into

https://github.com/IPPL-framework/ippl/commit/002b814aa8ca95f77e9d703935dd3fd4a19b4f73 Merge pull request https://github.com/IPPL-framework/ippl/pull/572 from IPPL-framework/pr501-pif-final
https://github.com/IPPL-framework/ippl/commit/758eca5114b0254745bcb702aa209fe3d378a694 Merge pull request https://github.com/IPPL-framework/ippl/pull/571 from IPPL-framework/pr501-nufft-final
https://github.com/IPPL-framework/ippl/commit/0424793e79ad7d7c55db0e8af6e238b31b23c836 Merge pull request https://github.com/IPPL-framework/ippl/pull/570 from IPPL-framework/pr501-fft-final
https://github.com/IPPL-framework/ippl/commit/1b66ed9e96df44adb0378559edbe9d96ebb27865 Merge pull request https://github.com/IPPL-framework/ippl/pull/569 from IPPL-framework/pr501-hosg-final
https://github.com/IPPL-framework/ippl/commit/73efeee5b6b2c372afd139794fe4c2f49497c7cf Merge pull request https://github.com/IPPL-framework/ippl/pull/544 from IPPL-framework/pr501-communication-particle-update
https://github.com/IPPL-framework/ippl/commit/b35146d51ef9eac836a38880afaf0e6cce433d9e Merge pull request https://github.com/IPPL-framework/ippl/pull/543 from IPPL-framework/pr501-pcg

and then merged.

PR #506, opened 2026-05-18
Add functionality to print fields in list format.

simon-schlepphorst:
For debugging, a feature to print fields in python list or C++ initializer_list compatible format is needed.

e.g. `[[0, 1], [2, 3]]`

This adds it.


src/Field/BareField.h
```diff
@@ -199,6 +199,16 @@ namespace ippl {
...
 
+        void write_as_list(std::ostream& out = std::cout) const;
+
+        void write_as_list(Inform& inf) const;
+
+        friend std::ostream& operator<<(std::ostream& out,
+                                        const BareField<T, Dim, ViewArgs...>& field) {
+            field.write_as_list(out);
+            return out;
+        }
```
aliemen, line 221: I know, it's more or less trivial. But a quick doxygen doc for these new functions would be nice. 

Please also mention that it only prints the rank local values, not the _whole_ field.

aliemen (changes requested):
Nice work, no comments on the print function. Only the `BareField` function perhaps needs some additional doc, see below. 

Please also add a quick unit test for the `os << BareField` operator you added. Just a small 2x2 field that you fill with some values, print and then compare the printed string to the expected string is probably enough. Can be appended to [`unit_tests/BareField/BareField.cpp`](https://github.com/IPPL-framework/ippl/blob/master/unit_tests/BareField/BareField.cpp).

src/Field/BareField.h
simon-schlepphorst: Done.

simon-schlepphorst:
Should be ready for the final review and merge.

unit_tests/BareField/BareField.cpp
```diff
@@ -259,6 +257,59 @@ TYPED_TEST(BareFieldTest, AllFuncs) {
...
+            }
+            if (layout.comm.rank() == 1) {
+                hview(0, 0) = 3;
+                hview(0, 1) = 4;
+            }
+        }
+
+        Kokkos::deep_copy(view, hview);
+
+        std::ostringstream ss;
```
aliemen: A `include <sstream>` would be nice, otherwise it only works because of a transitive include (`Ippl.h` includes `Inform.h`, which includes `sstream`).

unit_tests/BareField/BareField.cpp
```diff
@@ -259,6 +257,59 @@ TYPED_TEST(BareFieldTest, AllFuncs) {
...
+            if (layout.comm.rank() == 0) {
+                ASSERT_STREQ(ss.str().c_str(), "[[1, 2]]");
+            }
+            if (layout.comm.rank() == 1) {
+                ASSERT_STREQ(ss.str().c_str(), "[[3, 4]]");
+            }
+        } else
+            GTEST_SKIP();
+    } else
+        GTEST_SKIP();
```
aliemen: The "too many ranks skip" check should be at the beginning of the function, not at the end. Otherwise line 270 will try to construct the 2x2 field which is not possible e.g. on 4 ranks (it will run into [this error](https://github.com/IPPL-framework/ippl/blob/d1fde8ad21c148304babfac0769fa184ce94204c/src/FieldLayout/FieldLayout.hpp#L98)).

aliemen (changes requested):
Test looks good, but I believe the more-than-2-ranks skip is in the wrong place.

aliemen:

unit_tests/BareField/BareField.cpp
simon-schlepphorst: Sorry, the group having a coffee break in my office was a bit more distracting than anticipated. Won't happen again.

Should be fixed now.

simon-schlepphorst:

aliemen (approved):
Looks good now. There is a `std::cout` call that's technically not good in mpi programs, but not a problem here since tests are evaluated with the string stream anyways.

aaadelmann:
Something is failing here ...

aaadelmann (changes requested):
Tests must pass then I can approve

aliemen (changes requested):
I think I know what the error is, it's probably a GPU template bug and `ViewArgs` are not propagated correctly. On GH200 compilation I get the following errors:
- Incompatible View copy construction
- View assignment must have compatible spaces
- View assignment must have compatible layout or have rank <= 1

Basically when passing the `Kokkos::View<T**, Kokkos::LayoutRight, ExecSpace>` view to `BareField<T, Dim, ViewArgs...>::write_as_list`, it wants to call `detail::write_as_list<T, Dim>(dview_m, out);`, but the latter one resolves to/expects `const typename ViewType<T, Dim>::view_type& view`. 

...which is effectively `Kokkos ::View<T**>` with default layout/space. But `dview_m` is `Kokkos ::View<T**, Kokkos::LayoutRight, ExecSpace>`, so Kokkos tries an incompatible view conversion. 

I _think_ a quick fix is calling `detail::write_as_list<T, Dim, ViewArgs...>(dview_m, out);` instead. Alternatively, template `detail::write_as_list` directly on the view with e.g. `template <typename View>`.

simon-schlepphorst:
Templating directly on the view would be better, but that should be done consistently for all functions in `src/Utility/ViewUtils.h`.

aaadelmann (approved):

aliemen (approved):

PR #509, opened 2026-05-18
Fix particle archive memory spaces (fix all known failing tests)

### LINKED ISSUE #453: unit tests are segfaulting
Some unit tests, namely ParticleSendRecv, ORB, and PIC, are segfaulting when running them on 2 GPUs on Lumi. Right now they are in the FailingTests.cmake, but this should be fixed.

### LINKED ISSUE #508: ParticleSendRecv can fail on Device
# Problem

ParticleSendRecv can fail on Device builds during particle exchange with a device-side Kokkos inaccessible-memory-space assertion, followed by a CUDA deep-copy failure at the next synchronization point.

## Observed failure excerpt:

```Kokkos::View ERROR: ``` attempt to access inaccessible memory space (``` label="**UNAVAILABLE**" ```)
cudaErrorAssert: device-side assert triggered
Kokkos_CudaSpace.cpp:68

The failure appears during ParticleSendRecv.SendAndRecieve, after bunch->update() starts packing/sending particle attributes.

More details [here](https://my.cdash.org/tests/390330875)

## Current diagnosis
There are two likely memory-space issues in the particle send/receive path.

1. Test attribute memory space is inconsistent
In unit_tests/Particle/ParticleSendRecv.cpp, the particle layout is parameterized with ExecSpace:

```
using playout_type = ippl::ParticleSpatialLayout<T, Dim, mesh_type, ExecSpace>; 
```
expectedRank also uses that execution-space property:

using rank_type = ippl::ParticleAttrib<int, ExecSpace>;
but Q currently does not:

```
using charge_container_type = ippl::ParticleAttrib<T>;
```
On CUDA/HIP builds this can put attributes from the same particle container into different memory spaces. During ``` ParticleBase::sendToRank() ```, all attributes are packed and serialized. A CUDA kernel may then access an attribute or hash view through an inaccessible memory space.

2. Archive scalar serialization uses unmanaged byte views without preserving memory space
In src/Communicate/Archive.hpp, scalar serialize/deserialize creates unmanaged byte views as:
```
Kokkos::View<char*, Kokkos::MemoryUnmanaged> src_view(...);
Kokkos::View<char*, Kokkos::MemoryUnmanaged> dst_view(...);
```

These wrappers do not explicitly carry the source view memory space or archive buffer memory space. On device builds this can wrap a CUDA/HIP pointer as a host/default-space unmanaged view and then trigger the Kokkos inaccessible-memory-space assertion during ``` Kokkos::deep_copy ```.

## Proposed fix
Make the ParticleSendRecv test attribute Q use the same execution-space property as the particle layout:
```
using charge_container_type = ippl::ParticleAttrib<T, ExecSpace>;
```
In Archive.hpp, create unmanaged byte views with explicit memory spaces:

1. source view memory space for serialized attribute bytes
2. archive buffer memory space for archive bytes
3. reverse mapping for deserialization

A companion branch exists with this change: **particle-sendrecv-cuda-memory-space**
Commit: ec6f3c0a 

Fix particle archive memory spaces. 
Validated locally on the Serial build:

## Results:

- ParticleSendRecv passed single-rank locally
- ParticleSendRecv passed with 2 MPI ranks locally
- full local unit suite passed: 34/34

### PR DESCRIPTION
**Pull request by aaadelmann**
# Summary
This PR addresses a device failure observed in ParticleSendRecv.SendAndRecieve, where particle exchange can trigger a Kokkos inaccessible-memory-space assertion and is closing #508. 

In addition all known-failng test are now #453 running again! 

# Changes

This PR changes 

1. Q to:
```
using charge_container_type = ippl::ParticleAttrib<T, ExecSpace>;
```
2. Archive scalar byte views
```
Archive::serialize() and Archive::deserialize() 
```
wrapped raw byte pointers using:
```
Kokkos::View<char*, Kokkos::MemoryUnmanaged>
```
Those unmanaged byte views did not explicitly preserve the memory space of either the source attribute view or the archive buffer. On device builds this can wrap a CUDA/HIP pointer as a view in the wrong/default space and then fail during ``` Kokkos::deep_copy ```.

This PR changes the scalar archive byte views to use explicit memory spaces:

- source attribute view memory space for serialized source bytes
- archive buffer memory space for archive bytes
- reversed source/destination memory spaces for deserialization

# Changes
Make ParticleSendRecv::Bunch::Q use ParticleAttrib<T, ExecSpace>.
Preserve source and destination memory spaces in scalar Archive byte-copy paths.
Keep the vector archive path unchanged.

# Notes
The vector archive path still uses std::memcpy inside a Kokkos device lambda for Vector<T, Dim> serialization/deserialization. That path was not changed here because the reported failure stack points at the scalar archive/deep-copy path, and this PR keeps the fix narrowly scoped.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+ippl-IPPL-1.0.0
 build/
 doc/html
 doc/latex
--- a/cmake/FailingTests.cmake
+++ b/cmake/FailingTests.cmake
@@ -11,9 +11,6 @@
 if(BUILD_TESTING AND IPPL_MARK_FAILING_TESTS)
   set(IPPL_DISABLED_TEST_LIST
       AssembleRHS
-      ParticleSendRecv
-      ORB
-      PIC
       TestSolverDesign
       TestGaussian_convergence
       TestSphere
--- a/src/Communicate/Archive.hpp
+++ b/src/Communicate/Archive.hpp
@@ -26,8 +26,14 @@ namespace ippl {
             char* src_ptr         = (char*)(view.data());
             assert(writepos_m + (nsends * size) <= buffer_m.size());
             // construct temp views of the src/dst buffers of the correct size (bytes)
-            Kokkos::View<char*, Kokkos::MemoryUnmanaged> src_view(src_ptr, size * nsends);
-            Kokkos::View<char*, Kokkos::MemoryUnmanaged> dst_view(dst_ptr, size * nsends);
+            using src_view_type =
+                Kokkos::View<char*, typename Kokkos::View<T*, ViewArgs...>::memory_space,
+                             Kokkos::MemoryTraits<Kokkos::Unmanaged>>;
+            using dst_view_type =
+                Kokkos::View<char*, typename buffer_type::memory_space,
+                             Kokkos::MemoryTraits<Kokkos::Unmanaged>>;
+            src_view_type src_view(src_ptr, size * nsends);
+            dst_view_type dst_view(dst_ptr, size * nsends);
             Kokkos::deep_copy(dst_view, src_view);
             Kokkos::fence();
             writepos_m += (nsends * size);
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

File: .gitignore [hunk 1]
```diff
@@ -1,3 +1,4 @@
+ippl-IPPL-1.0.0
```
s-mayani, line 1: what is this in the gitignore? was it intentional?

s-mayani (approved):

.gitignore
aaadelmann: uups is a local thing ... but does not harm ...

Issue #513, opened 2026-05-19
Host/Device Constexpr Warning in Vector Reduction Identities


aaadelmann:
# Avoid CUDA Host/Device Constexpr Warning in Vector Reduction Identities

## Problem

When building IPPL with CUDA on GH200, `BareField` tests emit warning `#20013-D` from the `Kokkos::reduction_identity<ippl::Vector<T, Dim>>` specialization in `src/Field/BareField.hpp`.

The warning is triggered by calling `std::numeric_limits<T>::infinity()` inside `KOKKOS_FORCEINLINE_FUNCTION` methods:

```cpp
KOKKOS_FORCEINLINE_FUNCTION static ippl::Vector<T, Dim> min() {
    return ippl::Vector<T, Dim>(std::numeric_limits<T>::infinity());
}

KOKKOS_FORCEINLINE_FUNCTION static ippl::Vector<T, Dim> max() {
    return ippl::Vector<T, Dim>(-std::numeric_limits<T>::infinity());
}
```

NVCC diagnoses this because `std::numeric_limits<T>::infinity()` is treated as a constexpr host function being called from a host-device function.

We should avoid requiring the compiler workaround:

```text
--expt-relaxed-constexpr
```

## Proposed Fix

Delegate the scalar identity values back to Kokkos:

```cpp
KOKKOS_FORCEINLINE_FUNCTION static ippl::Vector<T, Dim> min() {
    return ippl::Vector<T, Dim>(Kokkos::reduction_identity<T>::min());
}

KOKKOS_FORCEINLINE_FUNCTION static ippl::Vector<T, Dim> max() {
    return ippl::Vector<T, Dim>(Kokkos::reduction_identity<T>::max());
}
```

This matches Kokkos reduction semantics and keeps vector field reductions consistent with scalar field reductions. For `float` and `double`, Kokkos uses finite extrema (`FLT_MAX`, `DBL_MAX`) rather than infinities, which avoids the CUDA host/device constexpr warning.

PR #514, opened 2026-05-19
Avoid numeric_limits in vector reduction identities

aaadelmann:
## Summary

This PR updates the `Kokkos::reduction_identity<ippl::Vector<T, Dim>>` specialization to use Kokkos' scalar reduction identities instead of `std::numeric_limits<T>::infinity()`.

The previous implementation called `std::numeric_limits<T>::infinity()` from `KOKKOS_FORCEINLINE_FUNCTION` methods:

The remaining warnings are not IPPL related, they come from 
            /user-environment/env/default/include/experimental/__p0009_bits/utility.hpp

```cpp
KOKKOS_FORCEINLINE_FUNCTION static ippl::Vector<T, Dim> min() {
    return ippl::Vector<T, Dim>(std::numeric_limits<T>::infinity());
}

KOKKOS_FORCEINLINE_FUNCTION static ippl::Vector<T, Dim> max() {
    return ippl::Vector<T, Dim>(-std::numeric_limits<T>::infinity());
}
```

NVCC diagnoses this on GH200 as warning `#20013-D`, because the `std::numeric_limits<T>::infinity()` call is treated as a constexpr host function called from a host-device function.

The new implementation delegates to Kokkos:

```cpp
KOKKOS_FORCEINLINE_FUNCTION static ippl::Vector<T, Dim> min() {
    return ippl::Vector<T, Dim>(Kokkos::reduction_identity<T>::min());
}

KOKKOS_FORCEINLINE_FUNCTION static ippl::Vector<T, Dim> max() {
    return ippl::Vector<T, Dim>(Kokkos::reduction_identity<T>::max());
}
```

## Changes

- Replace `std::numeric_limits<T>::infinity()` in vector min/max reduction identities with `Kokkos::reduction_identity<T>::min()` and `max()`.
- Remove the now-unused `<limits>` include from `src/Field/BareField.hpp`.
- closing #513

### DIFF
--- a/src/Field/BareField.hpp
+++ b/src/Field/BareField.hpp
@@ -6,7 +6,6 @@
 
 #include <Kokkos_ReductionIdentity.hpp>
 #include <cstdlib>
-#include <limits>
 #include <map>
 #include <utility>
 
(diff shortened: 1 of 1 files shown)


aliemen (approved):
Nice, looks good. 

Also I just noticed: `int` doesn't have an "infinity" so I am kind of curious if the reducers even worked for non-float types before...

PR #516, opened 2026-05-19
Avoid standard-library constructor paths in Kokkos device code

### LINKED ISSUE #515: `std::` Usage in Kokkos Device Code
## Problem

CUDA/NVCC can warn or fail when host-only C++ standard library functions are called from Kokkos device code. One concrete example was the warning from `std::numeric_limits<T>::infinity()` inside a `KOKKOS_FORCEINLINE_FUNCTION` reduction identity:

```text
warning #20013-D: calling a constexpr __host__ function from a __host__ __device__ function is not allowed
```

This was fixed by delegating to `Kokkos::reduction_identity<T>`, but the same class of issue may exist elsewhere when `std::` functions appear inside `KOKKOS_FUNCTION`, `KOKKOS_INLINE_FUNCTION`, `KOKKOS_LAMBDA`, or `KOKKOS_CLASS_LAMBDA` regions.

Not every `std::` use is problematic:

- Host-side setup code is fine.
- Type-only uses such as `std::size_t`, `std::int64_t`, and `std::index_sequence` are generally fine.
- `std::` calls after a `parallel_reduce` completes and before/after MPI communication are host-only and usually fine.

The risky cases are runtime `std::` function calls that are actually executed in device code.

## Initial Candidates

These occurrences appear to be inside Kokkos device-relevant code and should be reviewed:

- `src/Field/BareFieldOperations.hpp`: `std::pow(...)` inside a `KOKKOS_LAMBDA` used by `Field::norm`.
- `src/FEM/FEMVector.h`: `std::pow(...)` inside a `KOKKOS_LAMBDA` used by `FEMVector::norm`.
- `src/Particle/ParticleSpatialOverlapLayout.hpp`: `std::floor(...)` inside a `KOKKOS_INLINE_FUNCTION` used by particle cell indexing.

These can likely be replaced with Kokkos math equivalents:

```cpp
Kokkos::pow(...)
Kokkos::floor(...)
```
## Ambiguous Cases

Some constructors marked `KOKKOS_FUNCTION` use `std::initializer_list`, for example:

- `ippl::Vector`
- `ippl::NDIndex`
- `ippl::NDRegion`

These should be checked against CUDA/GH200 compiler output and actual device call paths before changing them. They may be harmless if only used on the host, but the annotation means they are candidates for closer inspection.

aaadelmann:
## Summary

This PR removes unnecessary `std::initializer_list` construction paths from small fixed-size types that are marked `KOKKOS_FUNCTION` / `KOKKOS_INLINE_FUNCTION`.

The affected types are:

- `ippl::Vector`
- `ippl::NDIndex`
- `ippl::NDRegion`

The goal is to make these constructors safer for CUDA/HIP device compilation and avoid relying on standard-library helper objects in code that may be instantiated for device execution.

## Changes

- Updated `Vector(args...)` to initialize its fixed storage directly instead of delegating through `Vector(std::initializer_list<T>)`.
- Added a size check to `Vector(std::initializer_list<T>)` using `Kokkos::abort` so malformed braced construction fails instead of silently leaving bad state or writing past the fixed array.
- Removed the private `NDIndex(std::initializer_list<Index>)` constructor.
- Constrained `NDIndex(args...)` to exactly `Dim` arguments and initialized `indices_m` directly.
- Replaced `NDIndex::length()`, `NDIndex::first()`, and `NDIndex::last()` implementations that used `std::index_sequence` with simple loops.
- Removed the private `NDRegion(std::initializer_list<PRegion<T>>)` constructor.
- Constrained `NDRegion(args...)` to exactly `Dim` arguments convertible to `PRegion<T>` and initialized `regions_m` directly.

## Rationale

These fixed-size types are used in Kokkos-callable code. Delegating variadic constructors through `std::initializer_list` is unnecessary and exposes device-callable constructors to standard-library objects that are not always robust under CUDA/HIP compilation.

Directly initializing the backing arrays keeps the behavior simpler, preserves compile-time dimension checking, and reduces the chance of host-only standard-library usage leaking into device instantiations.

## Commits

- `86c229fe` Make Vector constructors safer for device code
- `c01ef821` Avoid std constructs in NDIndex device code
- `3112ad78` Avoid initializer_list in NDRegion device constructor

## Notes

closes #515

### DIFF
--- a/src/Index/NDIndex.hpp
+++ b/src/Index/NDIndex.hpp
@@ -7,20 +7,9 @@
 
 namespace ippl {
     template <unsigned Dim>
-    template <class... Args>
+    template <class... Args, typename std::enable_if<sizeof...(Args) == Dim, bool>::type>
     KOKKOS_FUNCTION NDIndex<Dim>::NDIndex(const Args&... args)
-        : NDIndex({args...}) {
-        static_assert(Dim == sizeof...(args), "Wrong number of arguments.");
-    }
-
-    template <unsigned Dim>
-    KOKKOS_FUNCTION NDIndex<Dim>::NDIndex(std::initializer_list<Index> indices) {
-        unsigned int i = 0;
-        for (auto& index : indices) {
-            indices_m[i] = index;
-            ++i;
-        }
-    }
+        : indices_m{args...} {}
 
     template <unsigned Dim>
     KOKKOS_FUNCTION NDIndex<Dim>::NDIndex(const Vector<unsigned, Dim>& sizes) {
@@ -160,26 +149,29 @@ namespace ippl {
 
     template <unsigned Dim>
     KOKKOS_INLINE_FUNCTION Vector<size_t, Dim> NDIndex<Dim>::length() const {
-        auto construct = [&]<size_t... Idx>(const std::index_sequence<Idx...>&) {
-            return Vector<size_t, Dim>{indices_m[Idx].length()...};
-        };
-        return construct(std::make_index_sequence<Dim>{});
+        Vector<size_t, Dim> result;
+        for (unsigned d = 0; d < Dim; ++d) {
+            result[d] = indices_m[d].length();
+        }
+        return result;
     }
 
     template <unsigned Dim>
     KOKKOS_INLINE_FUNCTION Vector<int, Dim> NDIndex<Dim>::first() const {
-        auto construct = [&]<size_t... Idx>(const std::index_sequence<Idx...>&) {
-            return Vector<int, Dim>{indices_m[Idx].first()...};
-        };
-        return construct(std::make_index_sequence<Dim>{});
+        Vector<int, Dim> result;
+        for (unsigned d = 0; d < Dim; ++d) {
+            result[d] = indices_m[d].first();
+        }
+        return result;
     }
 
     template <unsigned Dim>
     KOKKOS_INLINE_FUNCTION Vector<int, Dim> NDIndex<Dim>::last() const {
-        auto construct = [&]<size_t... Idx>(const std::index_sequence<Idx...>&) {
-            return Vector<int, Dim>{indices_m[Idx].last()...};
-        };
-        return construct(std::make_index_sequence<Dim>{});
+        Vector<int, Dim> result;
+        for (unsigned d = 0; d < Dim; ++d) {
+            result[d] = indices_m[d].last();
+        }
+        return result;
     }
 
     template <unsigned Dim>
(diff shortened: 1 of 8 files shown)


src/Index/NDIndex.hpp
```diff
@@ -160,26 +149,29 @@ namespace ippl {
...
     KOKKOS_INLINE_FUNCTION Vector<size_t, Dim> NDIndex<Dim>::length() const {
-        auto construct = [&]<size_t... Idx>(const std::index_sequence<Idx...>&) {
-            return Vector<size_t, Dim>{indices_m[Idx].length()...};
-        };
-        return construct(std::make_index_sequence<Dim>{});
+        Vector<size_t, Dim> result;
+        for (unsigned d = 0; d < Dim; ++d) {
+            result[d] = indices_m[d].length();
+        }
+        return result;
```
aliemen, line 156: `Vector<size_t, Dim> result;` auto initializes every element to 0, see [here](https://github.com/IPPL-framework/ippl/blob/11957d9cfdd63280e6c97bd56c266230c27d7c7a/src/Types/Vector.h#L29). The previous `index_sequence` construction didn't. Meaning, every `NDIndex<Dim>::length()` call will 
1. initialize the vector to `0`.
2. overwrite every element again with `indices_m[d].length();`. 

This comment is for `lenght`/`first`/`last`. It might be a performance concern if used extensively (probably OK for `NDIndex`).

src/Types/Vector.hpp
```diff
@@ -34,7 +34,9 @@ namespace ippl {
 
     template <typename T, unsigned Dim>
     KOKKOS_FUNCTION Vector<T, Dim>::Vector(const std::initializer_list<T>& list) {
-        // PAssert(list.size() == Dim);
+        if (list.size() != Dim) {
+            Kokkos::abort("Vector initializer_list size must match Vector dimension");
+        }
```
aliemen: This will be called for every `Vector_t<T, Dim> v = {a, b, c};`-like call. Since those might be called billions of times per timestep (e.g. in OPALX), I don't think having a runtime check is a good idea here. 

I **think** (needs to be tested) it is completely fine to remove this one completely: the above `Vector<T, Dim>::Vector(const Args&... args)` signature will catch all "`{a, b, ...}`"-like initializations (when no explicit `initializer_list` object is used). Then we would have implicit `Dim` check at compile time.

aliemen (commented):
The syntax looks good. Since this changes initialization of `ippl::Vector`, I made some nit-picking comments regarding tiny changes in runtime performance. 

If you think that these are non-concerns, feel free to resolve my comments. I'll now try it on Daint with OPALX and leave another comment in a bit.

src/Types/Vector.hpp
aaadelmann: removed the check

aliemen (approved):

PR #518, opened 2026-05-19
Fix FEM quadrature lifetime in assembled RHS

### LINKED ISSUE #517: AssembleRHS is failing in debug mode
## Problem

```
Test project /Users/adelmann/git/ippl/build-kokkos-5.1.1-debug/unit_tests/FEM
    Start 13: AssembleRHS
1/1 Test #13: AssembleRHS ......................***Failed    0.11 sec
[==========] Running 6 tests from 6 test suites.
[----------] Global test environment set-up.
[----------] 1 test from AssembleRHSTest/0, where TypeParam = Parameters<double,Rank<1u>>
[ RUN      ] AssembleRHSTest/0.ConservationOfTotalWeight
Assertion failed: (this->ref_element_m.isPointInRefElement(localPoint) && "Point is not in reference element"), function evaluateRefElementShapeFunction, file LagrangeSpace.hpp, line 284.
[MacBook-Pro-66:94407] *** Process received signal ***
[MacBook-Pro-66:94407] Signal: Abort trap: 6 (6)
[MacBook-Pro-66:94407] Signal code:  (0)
[MacBook-Pro-66:94407] [ 0] 0   libsystem_platform.dylib            0x00000001888bb744 _sigtramp + 56
[MacBook-Pro-66:94407] [ 1] 0   libsystem_pthread.dylib             0x00000001888b1888 pthread_kill + 296
[MacBook-Pro-66:94407] [ 2] 0   libsystem_c.dylib                   0x00000001887b6850 abort + 124
```
In release mode the test passes. I use Kokkos 5.1.1. 

## Root Cause
Dangling references in FEM ownership. FiniteElementSpace stored `quadrature_m` as `const QuadratureType&`, while `unit_tests/FEM/AssembleRHS.cpp::make_space()` returns a Space built from a local quadrature. That leaves the returned space with a dangling quadrature reference. Release happened to pass because the invalid memory was not exposed; Debug hit the reference-element assertion.

## Resolution
I changed ownership instead of patching the test:

- src/FEM/FiniteElementSpace.h (line 229): quadrature_m is now stored by value.
- src/FEM/Quadrature/Quadrature.h (line 112): `ref_element_m` is now stored by value too, so copied quadrature objects do not keep dangling element references.

aaadelmann:
## Problem 
c.f. #517 
## Solution 
Store FEM quadrature and reference element objects by value instead of retaining references to caller-owned objects. AssembleRHS can construct a space from local element/quadrature temporaries, and the previous reference members left the returned space with dangling state. Debug builds exposed this as a reference-element assertion during lumped-mass assembly, while Release builds happened to run past the invalid lifetime.

## Note
If the solution causes problems, I can adapt the unit test.

## Closes #517

### DIFF
--- a/src/FEM/FiniteElementSpace.h
+++ b/src/FEM/FiniteElementSpace.h
@@ -226,7 +226,7 @@ namespace ippl {
 
         UniformCartesian<T, Dim>& mesh_m;
         ElementType ref_element_m;
-        const QuadratureType& quadrature_m;
+        QuadratureType quadrature_m;
         Vector<size_t, Dim> nr_m;
         Vector<double, Dim> hr_m;
         Vector<double, Dim> origin_m;
--- a/src/FEM/Quadrature/Quadrature.h
+++ b/src/FEM/Quadrature/Quadrature.h
@@ -109,7 +109,7 @@ namespace ippl {
 
     protected:
         unsigned degree_m;
-        const ElementType& ref_element_m;
+        ElementType ref_element_m;
         Vector<T, NumNodes1D> integration_nodes_m;
         Vector<T, NumNodes1D> weights_m;
 


s-mayani:
I am not sure this is right. The assembleRHS test is having issues due to how it constructs the space used in the test, as a quadrature object may outlive the reference element it has as reference. But in the code design itself for me it makes sense that the quadrature does not own an element, but needs it as a reference in order to create it (since we cannot have a quadrature without knowledge of the element). In the FiniteElementSpace class, to me it also makes sense to just have references to the quadrature and the element since a FiniteElementSpace does not own an element or a quadrature, but needs knowledge of them; however, the element had to be changed to be by value due to issues with access on device. The objects element, quadrature, and finite element space are created in the FEMPoissonSolver is the correct order such that the element is the one with the longest lifetime. Hence I think what needs to be fixed is the make_space in InterpolateDiracs, AssembleRHS, and AssembleCurrentRHS. Let me know if you agree. I guess in my argument the onus is placed on the user to ensure that the element is not destroyed before the quadrature.

aaadelmann:
Yes I think this is the better solution.

PR #519, opened 2026-05-20
Fix FEM test make_space lifetimes

s-mayani:
Fixes issue #517, proposing alternative solution to PR #518.

Can you check whether this fixed your issue @aaadelmann?


aaadelmann (approved):
tests do not fail anymore !

PR #523, opened 2026-05-24
Re-implement field write api from #506, but template view directly (no `auto` in function parameter)

aliemen:
See description of #506.

I want to re-run the jülrich ci and see if this was actually the problem. Here is what I changed. From this:
```cpp
template <unsigned Dim>
void write_as_list_impl(const auto view, std::ostream& out = std::cout) {
    auto N = view.extent(0);
    ...
```
to this:
```cpp
template <unsigned Dim, typename View>
void write_as_list_impl(const View& view, std::ostream& out = std::cout) {
    auto N = view.extent(0);
    ...
```
I think that the Jülich `cuda/12.6` + `gcc/13` combination didn't like this one. Newer `cuda` versions don't seem to have a problem with that (that's why the cscs ci passed).

### DIFF
--- a/src/Field/BareField.h
+++ b/src/Field/BareField.h
@@ -199,6 +199,27 @@ namespace ippl {
          */
         void write(Inform& inf) const;
 
+        /*!
+         * Print the rank local BareField.
+         * @param out stream
+         */
+        void write_as_list(std::ostream& out = std::cout) const;
+
+        /*!
+         * Print the rank local BareField.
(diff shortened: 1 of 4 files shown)


aliemen:
Looks like Jülich CI [passes now](https://ippl-bc4558.pages.jsc.fz-juelich.de/451798/allure-report/index.html)

srikrrish (approved):

PR #526, opened 2026-05-26
Fix unphysical results when load balancing and unit tests for odd ranks

s-mayani:
This PR addresses two issues:

- Unit tests failing for > 2 ranks in FEM. This exposed an issue in the way elements are distributed among ranks in initializeElementIndices(). The 4 rank tests are simply skipped since the domain cannot be divided into so many ranks, as it produces planes.

- The FEM solver + load balancing in Landau damping were giving incorrect results, as shown in the plot below. 
LandauDamping_Correctness_All_16_loadbalance

This is due to phi_m not being set to 0 before going into the solve after the updateLayout() call. This is fixed by adding a `*phi_m = 0` call in the LoadBalancer. The wrong results are fixed, as shown in the plot below:
LandauDamping_Correctness_All_16_loadbalance_fixed


alpine/LoadBalancer.hpp
```diff
@@ -61,7 +61,12 @@ class LoadBalancer {
 
         if (fs_m->getStype() == "CG" || fs_m->getStype() == "PCG" || fs_m->getStype() == "FEM" ||
             fs_m->getStype() == "FEM_PRECON") {
+            // The phi field needs to be updated and set to 0 in this case.
+            // This is because updateLayout() does NOT redistribute field values,
```
aaadelmann, line 65: Maybe you could make a link to #530

aaadelmann (changes requested):

alpine/LoadBalancer.hpp
s-mayani: Linked in the issue

aaadelmann (approved):

PR #528, opened 2026-05-26
Remove failing tests which don't fail anymore

s-mayani:


s-mayani:
The Debug build has one failing test on OpenMP, which is the LagrangeSpace. However, I do not see any output and cannot the test passes on my machine for Debug and Release builds.

aaadelmann (approved):

PR #538, opened 2026-06-02
check Issue #537

### LINKED ISSUE #537: LUMI cmake warnings
# CMake HIP configuration warnings on LUMI: `AMDGPU_TARGETS` not set / GPU detection fails

## Problem

When configuring IPPL with HIP on LUMI, CMake emits ROCm developer warnings because `AMDGPU_TARGETS` is not set and ROCm cannot auto-detect GPUs during configure.

Example:

```text
CMake Warning (dev) at /opt/rocm-6.3.4/lib/cmake/hip/hip-config-amd.cmake:91 (message):
   AMDGPU_TARGETS was not set, and system GPU detection was unsuccsesful.

   The amdgpu-arch tool failed:
   Error: 'Failed to get device count'
   Output: ''

   As a result, --offload-arch will not be set for subsuqent
   compilations, and the default architecture
   (gfx906 for dynamic build / gfx942 for static build) will be used
```

This can appear more than once, for example once through Kokkos/ROCTHRUST and once through HeFFTe/rocFFT.

On batch systems such as LUMI, CMake configure often runs in an environment where no GPU is visible. ROCm therefore cannot infer the AMD GPU architecture using amdgpu-arch.

If AMDGPU_TARGETS is not set explicitly, ROCm may fall back to a default architecture that does not match the target hardware. On LUMI-G the intended target is typically gfx90a.

## Suggested Fix

```text 
if(IPPL_ENABLE_HIP OR Kokkos_ENABLE_HIP)
    if(DEFINED CMAKE_HIP_ARCHITECTURES AND NOT DEFINED AMDGPU_TARGETS)
        set(AMDGPU_TARGETS "${CMAKE_HIP_ARCHITECTURES}" CACHE STRING
            "AMDGPU targets for ROCm packages" FORCE)
    endif()
endif()
```
This should be placed early enough that Kokkos, rocFFT, rocPRIM, rocThrust, and HeFFTe see the value before calling find_package() on ROCm components.

aaadelmann:
Fix some warning at AMD (LUMI) details in #537 

Close #537


aaadelmann:
cscs-ci run  cscs-ci-mi300

srikrrish (approved):

PR #543, opened 2026-06-10
Pr501 pcg

aaadelmann:
# PCG/Preconditioner

## Summary

This PR extracts the PCG allocation/performance part from the larger PR501 work.
It keeps the scope intentionally small:

- reuse CG/PCG work fields instead of constructing them inside every solve,
- make preconditioners write into caller-provided output fields,
- pass solver operator fields by reference instead of by value,
- keep preconditioner scratch fields synchronized with layout changes,
- fix ALPINE particle initialization code that took pointers/references to temporary Kokkos views.

The primary motivation is to remove repeated allocation and avoid repeated
halo-buffer growth on copied fields in the PCG/preconditioner path. This was
observed as repeated allocation work in multi-rank profiling.

### PCG workspaces are reused

`src/LinearSolvers/PCG.h`

 In the CG/PCG algorithm these are:
  - `r`: the current residual, initially `rhs - A * lhs`,
  - `d`: the current search direction,
  - `q`: the operator-applied search direction, i.e. `A * d`.
 
- PCG adds reusable `s` and `pcond_out` buffers.
  Here `s` stores the preconditioned residual `M^{-1} r`, and `pcond_out`
  stores the initial preconditioner output before it is folded into `d`.

- `PCG::operator()` refreshes workspace layouts instead of constructing local
  fields on every solve.
- `op_m(d)` now assigns into the existing `q` storage instead of forcing a
  `deepCopy()`.
- Preconditioner applications now write into `pcond_out` / `s` instead of
  returning freshly allocated fields.

Important BC detail:

- `pcond_out` and `s` intentionally keep their default `NoBcFace` boundary
  conditions. This preserves the old behavior where the preconditioner returned
  a fresh field without periodic BCs. If these buffers inherited periodic BCs,
  preconditioner-internal operators could trigger extra `PeriodicFace::apply`
  MPI calls and diverge from the expected multi-rank communication sequence.

### Preconditioners now write into caller-owned output

`src/LinearSolvers/Preconditioner.h`

- The base preconditioner interface changed from returning `Field` to:

  ```cpp
  void operator()(Field& input, Field& result)
  ```

- Concrete preconditioners now reuse member scratch fields where possible.
- `init_fields()` allocates scratch fields once and updates their layout on
  subsequent calls.
- Several preconditioners still keep necessary `deepCopy()` calls where
  expression aliasing can occur, especially for field-valued inverse-diagonal
  operators.

Covered preconditioners include:

- identity/default,
- Jacobi,
- polynomial Newton,
- polynomial Chebyshev,
- Richardson,
- alternative Richardson,
- Gauss-Seidel,
- SSOR.

### Solver operators take fields by reference

`src/LinearSolvers/PCG.h`
`src/LinearSolvers/Preconditioner.h`
`src/PoissonSolvers/PoissonCG.h`

The solver operator wrapper now takes the field by reference:

```cpp
#define IPPL_SOLVER_OPERATOR_WRAPPER(fun, type) \
    [](type& arg) {                             \
        return fun(arg);                        \
    }
```

This matters because a by-value `Field` copy shares the underlying data views but
does not share all mutable side state, notably halo exchange buffers that may be
grown through `Kokkos::realloc`. When the operator acts on a copied field, the
copy can grow its halo buffer without updating the original. In multi-rank runs
this showed up as repeated allocation work.

`PoissonCG.h` no longer redefines the wrapper locally. Keeping a single
by-reference wrapper avoids silently reintroducing the by-value behavior.

### ALPINE Kokkos view lifetime fixes

The ALPINE managers no longer take addresses of temporary view handles returned
by `getView()`.

Changed files:

- `alpine/LandauDampingManager.h`
- `alpine/BumponTailInstabilityManager.h`
- `alpine/PenningTrapManager.h`

Before:

```cpp
view_type* R = &(this->pcontainer_m->R.getView());
samplingR.generate(*R, rand_pool64);
```

After:

```cpp
view_type R = this->pcontainer_m->R.getView();
samplingR.generate(R, rand_pool64);
```

This fixes compilers/backends that reject taking the address of a temporary
Kokkos view object and avoids dangling view-handle pointers.

## Documentation Check

The implementation is documented in the places where review context matters:

- `Preconditioner.h` documents why the operator wrapper must take fields by
  reference.
- `PCG.h` documents why workspaces are reused and why `pcond_out` / `s` keep
  `NoBcFace` boundary conditions.
- `PoissonCG.h` documents why the local wrapper definition was removed.
- `PR501_SPLIT_MAP.md` documents the split rationale and dependencies between
  the planned PR501 follow-up branches.

No user-facing manual update appears necessary because this PR changes internal
solver allocation behavior and preserves the public PCG/Poisson solver behavior.

### Original split-map validation

The PCG split prototype was built and tested with unit tests:

```text
100% tests passed, 0 tests failed out of 34
Total Test time (real) = 54.17 sec
```

### Solver integration validation

A dedicated solver integration build was configured:

```bash
cmake -S . -B build-pr501-pcg-ssor-tests-release \
  -DCMAKE_BUILD_TYPE=Release \
  -DIPPL_PLATFORMS=SERIAL \
  -DIPPL_ENABLE_TESTS=ON \
  -DIPPL_ENABLE_UNIT_TESTS=OFF \
  -DIPPL_ENABLE_SOLVERS=ON \
  -DIPPL_ENABLE_FFT=OFF
```

Built targets:

```bash
cmake --build build-pr501-pcg-ssor-tests-release \
  --target TestCGSolver TestCGSolver_convergence TestSolverDesign -j 8
```

Results:

- `TestCGSolver`: built and passed through CTest.
- `TestCGSolver_convergence`: compile-only target built successfully.
- `TestSolverDesign`: compile-only target built successfully.

Registered CTest:

```bash
ctest --test-dir build-pr501-pcg-ssor-tests-release \
  --output-on-failure -R '^TestCGSolver$'
```

Result:

```text
TestCGSolver passed
```

The SSOR path was also run explicitly because the registered CTest uses Jacobi:

```bash
./test/solver/TestCGSolver 4 s 2 1 1.57079632679 --info 5
```

and with MPI:

```bash
/opt/homebrew/Cellar/open-mpi/5.0.8/bin/mpiexec -n 2 \
  ./test/solver/TestCGSolver 4 s 2 1 1.57079632679 --info 5

/opt/homebrew/Cellar/open-mpi/5.0.8/bin/mpiexec -n 4 \
  ./test/solver/TestCGSolver 4 s 2 1 1.57079632679 --info 5
```

Results:

| ranks | residual | iterations |
|---:|---:|---:|
| 1 | `1.514653902775893e-13` | 42 |
| 2 | `1.196970356284836e-13` | 44 |
| 4 | `1.588802017354866e-13` | 38 |

The same explicit SSOR test was temporarily rerun with the preconditioner
changes reverted. Residuals and iteration counts were identical for 1, 2, and 4
ranks.

## Reviewer Notes

- The key semantic change is the preconditioner API: preconditioners now write
  into a supplied output field rather than returning a new field.
- The by-reference solver operator wrapper is intentional and important for halo
  buffer reuse.
- `pcond_out` and `s` should not inherit periodic BCs; that would change the MPI
  call pattern inside preconditioner operator chains.
- Some `deepCopy()` calls remain intentionally where expression aliasing can
  otherwise corrupt field-valued preconditioner operations.


s-mayani:
Two things I don't understand:

- "fix ALPINE particle initialization code that took pointers/references to temporary Kokkos views."
- "pcond_out and s intentionally keep their default NoBcFace boundary
conditions. This preserves the old behavior where the preconditioner returned
a fresh field without periodic BCs. If these buffers inherited periodic BCs,
preconditioner-internal operators could trigger extra PeriodicFace::apply
MPI calls and diverge from the expected multi-rank communication sequence."

For the first one, I don't think I understand why this change is necessary. @srikrrish do you know?

For the second one, I am a bit confused by the wording, and I also think the comment in the code regarding this could be made more informative for future developers, especially considering (if I understood correctly) that this was the source of deadlocks in PCG.

alpine/BumponTailInstabilityManager.h
```diff
@@ -274,10 +274,10 @@ class BumponTailInstabilityManager : public AlpineManager<T, Dim> {
 
         this->pcontainer_m->create(nlocal);
 
-        view_type* R = &(this->pcontainer_m->R.getView());
-        samplingR.generate(*R, rand_pool64);
+        view_type R = this->pcontainer_m->R.getView();
+        samplingR.generate(R, rand_pool64);
 
-        view_type* P = &(this->pcontainer_m->P.getView());
+        view_type P = this->pcontainer_m->P.getView();
```
s-mayani, line 280: why was this needed?

s-mayani: oh is it because a view is anyways a pointer?

src/LinearSolvers/PCG.h
```diff
@@ -327,18 +330,37 @@ namespace ippl {
...
             , preconditioner_m(nullptr){};
 
+        /*
+         * Extends the CG workspace with the preconditioner result buffers s and
+         * pcond_out so operator() does not allocate per solve. pcond_out keeps
+         * its default NoBcFace BCs: the preconditioner's internal operator
+         * chain therefore never triggers PeriodicFace::apply MPI calls -- if it
+         * did, the global MPI sequence would diverge from the master code path
+         * (where the preconditioner returned a fresh NoBcFace field) and
+         * intermittent multi-rank halo deadlocks would follow.
```
s-mayani: I do not understand this comment... I think it could be made more informative.

src/LinearSolvers/PCG.h
```diff
@@ -422,19 +444,26 @@ namespace ippl {
...
-            lhs_type d(mesh, layout);
-            lhs_type s(mesh, layout);
-            lhs_type q(mesh, layout);
-
+            // r, d, q come from the CG base class; s and pcond_out are PCG
+            // members. All are pre-allocated via initializeFields(); operator()
+            // only refreshes their layout so a load-balance repartition is
+            // tracked. The preconditioner scratch must follow the lhs layout
+            // too, otherwise its halo-exchange neighbor list goes out of sync
+            // with r/d/s/q after a repartition and halo MPI calls deadlock.
```
s-mayani: too verbose, can change to
"Field layouts are updated such that we don't keep a stale domain decomposition if the lhs layout has been changed during the simulation, for example by the load balancer".

src/LinearSolvers/PCG.h
```diff
@@ -422,19 +444,26 @@ namespace ippl {
...
+            this->r.updateLayout(lhs.getLayout());
+            this->d.updateLayout(lhs.getLayout());
+            s.updateLayout(lhs.getLayout());
+            pcond_out.updateLayout(lhs.getLayout());
+            this->q.updateLayout(lhs.getLayout());
+
+            // Each preconditioner's init_fields() is responsible for being
+            // cheap on the steady-state path (refreshing layout in-place, not
+            // reallocating).
```
s-mayani, line 445: I don't understand this comment.  Isn't the "refreshing layout in-place" is done by the lines of code above (updateLayout)? What does steady-state path mean?

src/LinearSolvers/PCG.h
```diff
@@ -459,20 +488,26 @@ namespace ippl {
...
             }
 
-            r = rhs - this->op_m(lhs);
-            d = preconditioner_m->operator()(r).deepCopy();
-            d.setFieldBC(bc);
+            this->r = rhs - this->op_m(lhs);
+            // pcond_out keeps its default NoBcFace BCs so the preconditioner's
+            // internal operator chain does not trigger PeriodicFace::apply MPI;
+            // the expression assignment then folds it into d (which gets the
+            // periodic BCs via setFieldBC below).
```
s-mayani: I don't understand this comment

src/LinearSolvers/PCG.h
```diff
@@ -459,20 +488,26 @@ namespace ippl {
...
             this->residueNorm = Kokkos::sqrt(Kokkos::abs(delta1));
             const T tolerance = params.get<T>("tolerance") * this->residueNorm;
 
             while (this->iterations_m < maxIterations && this->residueNorm > tolerance) {
-                q       = this->op_m(d);
-                q       = q.deepCopy();
-                T alpha = delta1 / innerProduct(d, q);
-                lhs     = lhs + alpha * d;
+                // op_m(d) writes its expression into q's existing storage; no
+                // allocation, no per-iteration deep copy.
```
s-mayani, line 482: this comment is not needed

src/LinearSolvers/PCG.h
```diff
@@ -502,6 +538,13 @@ namespace ippl {
 
     protected:
         std::unique_ptr<preconditioner<FieldLHS>> preconditioner_m;
+
+        // Preconditioner result buffers, allocated once via initializeFields()
+        // and reused across solves. Both deliberately keep their default
+        // NoBcFace BCs so the preconditioner's internal operator chain does
+        // NOT trigger PeriodicFace::apply MPI calls.
```
s-mayani: again I am not sure I understand this

src/LinearSolvers/Preconditioner.h
```diff
@@ -125,12 +145,64 @@ namespace ippl {
...
+            recursive_preconditioner(u, level - 1, Pr);
+            IpplTimings::startTimer(t_op);
+            PA = op_m(Pr);
+            IpplTimings::stopTimer(t_op);
+            recursive_preconditioner(PA, level - 1, PAPr);
+            IpplTimings::startTimer(t_comb);
+            out = eta_m[level] * (2.0 * Pr - PAPr);
+            IpplTimings::stopTimer(t_comb);
+            IpplTimings::stopTimer(t_total);
+        }
```
s-mayani, line 177: I think all these changes are not necessary, I am not even sure why this recursive preconditioner was touched.

src/LinearSolvers/Preconditioner.h
```diff
@@ -125,12 +145,64 @@ namespace ippl {
...
+
+        void operator()(Field& u, Field& result) override {
+            // One outer timer per polynomial_newton apply, complementing the
+            // per-call recursion timers. Lets us see how many outer pcond
+            // calls a solve does vs. the time inside each.
+            static IpplTimings::TimerRef t_apply = IpplTimings::getTimer("newton/apply");
+            IpplTimings::startTimer(t_apply);
+            recursive_preconditioner(u, level_m, result);
+            IpplTimings::stopTimer(t_apply);
```
s-mayani: Same for all these extra comments and timers, not necessary changes, and this preconditioner is never used.

src/LinearSolvers/Preconditioner.h
```diff
@@ -125,12 +145,64 @@ namespace ippl {
...
+        }
+
+        void init_fields(Field& b) override {
+            // One-shot precomputation of the eta coefficients.
```
s-mayani, line 187: why was this comment changed, it was already good before "Precompute the etas for later use".

src/LinearSolvers/Preconditioner.h
```diff
@@ -170,6 +248,10 @@ namespace ippl {
         double zeta_m;  // smallest (alpha + beta) is multiplied by (1+zeta) to avoid clustering of
                         // Eigenvalues
         double* eta_m = nullptr;  // Size is determined at runtime
+        std::vector<Field> Pr_scratch_m;
+        std::vector<Field> PA_scratch_m;
+        std::vector<Field> PAPr_scratch_m;
+        bool fields_initialized_m = false;
```
s-mayani, line 237: Isn't it super expensive to store all these temporary fields which are not needed later (needed only for the recursion)?

src/LinearSolvers/Preconditioner.h
```diff
@@ -219,19 +301,65 @@ namespace ippl {
...
+                // x_old gets a deep copy of the previous x.
+                result = rho_m[i] * (2 * sigma_m * x_m - rho_m[i - 1] * x_old_m + z_m);
+                IpplTimings::stopTimer(t_combo);
+                IpplTimings::startTimer(t_copy);
+                Kokkos::deep_copy(x_old_m.getView(), x_m.getView());
+                Kokkos::deep_copy(x_m.getView(), result.getView());
+                IpplTimings::stopTimer(t_copy);
+            }
+            IpplTimings::stopTimer(t_apply);
+        }
```
s-mayani, line 313: All these extra timers are not needed in this preconditioner, since we never use it, and the extra comments also don't seems necessary to me.
It makes the algorithm cumbersome to read.

src/LinearSolvers/Preconditioner.h
```diff
@@ -295,52 +422,50 @@ namespace ippl {
...
-            // memory at every iteration.
-            // In order for the operator calls to not rewrite
-            // on this same field over and over when calling 
-            // the operators (upper, diag, inverse, lower, etc)
-            // we need deep copies to the preconditioner fields.
+        void operator()(Field& r, Field& result) override {
+            // Richardson iteration in-place on the caller-provided result
+            // buffer. ULg_m stays as a member scratch; the inner deep copies
+            // remain because the (upper, diag, inverse, lower) operators may
+            // return Field-valued expressions that alias their input.
```
s-mayani, line 384: Can we change the comment to "Richardson iteration in-place (using the residual r) on the caller-provided result field.
In order for the operator calls to not rewrite on this same field over and over when calling the operators (upper, diag, inverse, lower, etc) we need deep copies to the preconditioner fields."

alpine/BumponTailInstabilityManager.h
aaadelmann: Yes and see also the comment in the PR under **ALPINE Kokkos view lifetime fixes**

src/LinearSolvers/Preconditioner.h
```diff
@@ -435,73 +560,55 @@ namespace ippl {
...
         }
 
-        Field operator()(Field& b) override {
-            layout_type& layout = b.getLayout();
-            mesh_type& mesh     = b.get_mesh();
+        void operator()(Field& b, Field& result) override {
+            // The running iterate lives in result; UL_m and r_m are scratch
+            // members. The inner deep copies remain because the (upper, lower,
+            // inverse) operators may return Field-valued expressions that
+            // alias their input.
```
s-mayani, line 522: Again same change: "In order for the operator calls to not rewrite on this same field over and over when calling the operators (upper, diag, inverse, lower, etc) we need deep copies to the preconditioner fields."

src/LinearSolvers/Preconditioner.h
```diff
@@ -364,22 +489,19 @@ namespace ippl {
...
-            // memory at every iteration.
-            // In order for the operator calls to not rewrite
-            // on this same field over and over when calling 
-            // the operators (upper, diag, inverse, lower, etc)
-            // we need deep copies to the preconditioner fields.
+        void operator()(Field& r, Field& result) override {
+            // result holds the running iterate; Ag_m and g_old_m are scratch
+            // members. The inner deep copies remain because the operators
+            // (op_m, inverse_diagonal_m) may return Field-valued expressions
+            // that alias their input.
```
s-mayani, line 451: Same change:
"In order for the operator calls to not rewrite on this same field over and over when calling the operators (upper, diag, inverse, lower, etc) we need deep copies to the preconditioner fields."

src/LinearSolvers/Preconditioner.h
```diff
@@ -435,73 +560,55 @@ namespace ippl {
...
+                    UL_m = lower_m(result);
                     UL_m = UL_m.deepCopy();
-                    x    = r_m - UL_m;
-                    // The inverse diagonal is applied to the
-                    // vector itself to return the result usually.
-                    // However, the operator for FEM already
-                    // returns the result of inv_diag * itself
-                    // due to the matrix-free evaluation.
-                    // Therefore, we need this if to differentiate
-                    // the two cases.
```
s-mayani, line 468: why was this comment removed?

src/LinearSolvers/Preconditioner.h
```diff
@@ -435,73 +560,55 @@ namespace ippl {
...
+                    UL_m = upper_m(result);
                     UL_m = UL_m.deepCopy();
-                    x    = r_m - UL_m;
-                    // The inverse diagonal is applied to the
-                    // vector itself to return the result usually.
-                    // However, the operator for FEM already
-                    // returns the result of inv_diag * itself
-                    // due to the matrix-free evaluation.
-                    // Therefore, we need this if to differentiate
-                    // the two cases.
```
s-mayani, line 488: Why was this comment removed?

src/LinearSolvers/PCG.h
aaadelmann: Yes maybe to verbose for the specialist but I personally like it and furthermore students maybe appreciate it very much

src/LinearSolvers/Preconditioner.h
```diff
@@ -536,86 +644,73 @@ namespace ippl {
...
 
-            layout_type& layout = b.getLayout();
-            mesh_type& mesh     = b.get_mesh();
-
-            Field x(mesh, layout);
-
-            x = 0;  // Initial guess
+            // The running iterate lives in result; UL_m, r_m are scratch
+            // members. The inner deep copies remain because (upper, lower,
+            // inverse, diagonal) may return aliasing Field-valued expressions.
```
s-mayani: Again same change: "In order for the operator calls to not rewrite on this same field over and over when calling the operators (upper, diag, inverse, lower, etc) we need deep copies to the preconditioner fields."

src/LinearSolvers/Preconditioner.h
```diff
@@ -536,86 +644,73 @@ namespace ippl {
...
-            // the operators (upper, diag, inverse, lower, etc)
-            // we need deep copies to the preconditioner fields.
-
-            // The inverse diagonal is applied to the
-            // vector itself to return the result usually.
-            // However, the operator for FEM already
-            // returns the result of inv_diag * itself
-            // due to the matrix-free evaluation.
-            // Therefore, we need this if to differentiate
-            // the two cases.
```
s-mayani, line 571: Why was this comment removed?

src/PoissonSolvers/PoissonCG.h
```diff
@@ -11,13 +11,12 @@
...
-#define IPPL_SOLVER_OPERATOR_WRAPPER(fun, type) \
-    [](type arg) {                              \
-        return fun(arg);                        \
-    }
+// IPPL_SOLVER_OPERATOR_WRAPPER is defined once in LinearSolvers/Preconditioner.h
+// (re-exported through this header via PCG.h). Defining it again here used to
+// silently shadow that definition with a by-value lambda, which copies the
+// Field on every op_m() call and reintroduces the per-iteration cudaMalloc
+// in the halo exchange (the realloc never propagates back to the original
+// Field). Keep a single by-reference definition.
```
s-mayani, line 19: If it is removed from the new file, we can also remove this comment, it is not needed.

PR501_SPLIT_MAP.md
s-mayani, line 1: This file should not be part of the PR.

s-mayani (changes requested):

srikrrish:
> This fixes compilers/backends that reject taking the address of a temporary Kokkos view object and avoids dangling view-handle pointers.

I think it is due to the explanation written here 
"This fixes compilers/backends that reject taking the address of a temporary
Kokkos view object and avoids dangling view-handle pointers." 
I am not sure if these things were observed before or it is just a pre-cautionary measure.

s-mayani:
> > This fixes compilers/backends that reject taking the address of a temporary Kokkos view object and avoids dangling view-handle pointers.
> 
> I think it is due to the explanation written here "This fixes compilers/backends that reject taking the address of a temporary Kokkos view object and avoids dangling view-handle pointers." I am not sure if these things were observed before or it is just a pre-cautionary measure.

Okay thanks!

s-mayani:
In general, I think I disagree with the use of "result buffer" in most of the comments regarding the fields stored in CG and PCG. It might create a confusion with the communication buffers, "results field" would be less ambiguous.

aaadelmann:
Note: the comments I have checked random are the original ones, not some Hugo hallucination

aaadelmann:
I think I checked on all of the items but not sure because git not allowing me to resolve some of them ... pls. check

src/LinearSolvers/PCG.h
```diff
@@ -502,7 +477,7 @@ namespace ippl {
             this->residueNorm = Kokkos::sqrt(Kokkos::abs(delta1));
             const T tolerance = params.get<T>("tolerance") * this->residueNorm;
 
-            while (this->iterations_m < maxIterations && this->residueNorm > tolerance) {
+            while (this->iterations_m<maxIterations&& this->residueNorm> tolerance) {
```
s-mayani, line 480: why are the spaces removed here?

s-mayani (dismissed):

PR501_SPLIT_MAP.md
s-mayani, line 1: this should be removed from the PR

s-mayani (approved):

PR #545, opened 2026-06-11
Fixing CI on beverin

biddisco:


biddisco:
ci-cscs run cscs-ci-mi300

PR #554, opened 2026-06-20
Multigrid Preconditioner

NickvonWit:
# Multigrid Preconditioner + Strided-Index Support

## What's added

**`src/LinearSolvers/Multigrid.h`** (new) — Matrix-free geometric multigrid V-cycle preconditioner (`multigrid_preconditioner<Field, OperatorF>`), deriving from `preconditioner<Field>`:
- The `setDebugPrint`/`printDebugState`/`debug_print_*` helpers and the commented-out instrumented `vcycle` are kept intentionally for future debugging.

**Solver wiring:**
- `PCG.h` — added `multigrid` to `valid_types`, added MG defaults (`mg_pre_smooth=2`, `mg_post_smooth=2`, `mg_omega=0.8`, `mg_min_cells=4`, `mg_communication=false`), threaded the MG params through all `setPreconditioner` signatures, and added a `multigrid` construction branch.
- `PoissonCG.h` — reads MG params via optional defaulted lookups so legacy solvers without those keys still work; passes them through both operator-wrapper branches.
- `FieldSolver.hpp` — `ParsedPreconditionerParams` gains 4 MG fields, the parser accepts `multigrid` (4 args), and CG setup adds the MG keys to the `ParameterList`.

**Tests** — new `test/solver/multigrid/` subdir (registered in `test/CMakeLists.txt`): `TestMultigrid` (runtime) plus 13 `COMPILE_ONLY` tests (discrete/convergence/tolerance/comparison × constant/periodic/anisotropic). Also new `test/solver/TestCGSolver_convergence.cpp` and MG params added to `TestCGSolver.cpp`.

`.gitignore` adds `compile_commands.json`.

## IPPL striding fixes

Previously the code assumed `stride == 1`. Coarse multigrid levels use strided indices, so periodic offsets and local↔view index conversions must account for the stride.

| File | Function | Fix |
|---|---|---|
| `src/Index/Index.hpp` | `Index::grow` | `first_m -= ncells * stride_m` (was `- ncells`) |
| `src/FieldLayout/FieldLayout.hpp` | `getPeriodicOffset` | offset uses `length() * stride()` instead of `length()` |
| `src/FieldLayout/FieldLayout.hpp` | intersect-range loop | `lo/hi = (overlap - offset.first()) / stride + nghost (+1)` |
| `src/Field/BcTypes.hpp` | periodic BC (serial + parallel) | `period = length() * stride()`; receive-range divided by `stride` |
| `src/Field/HaloCells.hpp` | send/recv boundary detection | compute `globalLo/globalHi` with `* stride` before comparing to `domain.min()/max()` |

**`src/FieldLayout/SubFieldLayout.hpp`** — behavioural change: the local-domain restriction now unconditionally intersects every rank's local domain with the sub-domain (removed the `if (!gDomain_m.contains(...))` guard).

## clang-format-only changes

No logic changes in these files:

- `alpine/AlpineManager.h`
- `alpine/FieldContainer.hpp`
- `alpine/FieldSolver.hpp`
- `alpine/LandauDamping.cpp`
- `alpine/LandauDampingManager.h`
- `src/LinearSolvers/PCG.h`
- `src/PoissonSolvers/PoissonCG.h`
- `src/FieldLayout/SubFieldLayout.hpp`
- `test/solver/TestCGSolver.cpp`

The `owned` → `global_domain` rename in `TestFieldBC`, `TestHalo`, `TestHessian`, `TestLaplace`, and `TestMixedExec` is a rename only, no logic change.


aaadelmann:

src/LinearSolvers/Multigrid.h
```diff
@@ -0,0 +1,669 @@
...
+                                 unsigned min_cells_per_rank_per_dim = 4, bool communication = true)
+            : preconditioner<Field>("Multigrid")
+            , op_(std::forward<OperatorF>(op))
+            , nu1_(pre_smooth_iters)
+            , nu2_(post_smooth_iters)
+            , omega_(omega_jacobi)
+            , min_cells_per_rank_per_dim_(min_cells_per_rank_per_dim)
+            , communication_(communication) {}
+
+        // --- DEBUGGING - To be deleted ---
```
s-mayani: if the debugging print functions are going to be kept, please remove the "To be deleted" comment

src/LinearSolvers/Multigrid.h
```diff
@@ -0,0 +1,669 @@
...
+
+            L_[0].f = b.deepCopy();
+
+            // Remove Volume average if periodic
+            if (is_all_periodic_) {
+                auto avg = L_[0].f.getVolumeAverage();
+                L_[0].f  = L_[0].f - avg;
+                if (communication_)
+                    L_[0].f.fillHalo();
+            }
```
s-mayani, line 184: why do we need to do this for f? shouldn't this only be needed for the solution u?

src/LinearSolvers/Multigrid.h
```diff
@@ -0,0 +1,669 @@
...
+                L_[0].u  = L_[0].u - avg;
+                if (communication_)
+                    L_[0].u.fillHalo();
+            }
+
+            // Write the result into the caller-provided buffer instead of
+            // returning by value. deep_copy of the underlying views keeps
+            // result's own BCs/halo metadata intact (matching the identity
+            // and chebyshev preconditioners), so the PCG outer operator chain
+            // does not inherit L_[0].u's field BCs.
```
s-mayani: why is this desirable?

src/LinearSolvers/Multigrid.h
```diff
@@ -0,0 +1,669 @@
...
+            auto& fine_mesh              = b.get_mesh();
+            auto& fine_layout            = b.getLayout();
+            auto fine_domain             = fine_layout.getDomain();
+            std::array<bool, Dim> decomp = fine_layout.isParallel();
+
+            auto bcs = b.getFieldBC();
+
+            // 1. Calculate number of levels
+            const auto& localFine = fine_layout.getLocalNDIndex();  // this rank's slab
+            int min_local         = std::numeric_limits<int>::max();
```
s-mayani: can we replace it with a Kokkos:: max?

src/LinearSolvers/Multigrid.h
```diff
@@ -0,0 +1,669 @@
...
+
+            for (unsigned d = 0; d < Dim; ++d) {
+                // For serial dims, every rank's local extent == global extent, so this naturally
+                // covers them too. We do not have to check if dim is parallel or not.
+                min_local = std::min(min_local, static_cast<int>(localFine[d].length()));
```
s-mayani: same here, is it possible to replace std:: with Kokkos:: math functions everywhere?

src/LinearSolvers/Multigrid.h
```diff
@@ -0,0 +1,669 @@
...
+                    apply(fc, args) = sum / denom;
+                });
+
+            ippl::fence();
+
+            // Remove Volume average if periodic
+            if (is_all_periodic_) {
+                auto avg     = lev_coarse.f.getVolumeAverage();
+                lev_coarse.f = lev_coarse.f - avg;
+            }
```
s-mayani, line 377: if we are already removing the volume average of f at each level, is it needed in the operator()?
also why do you need to remove it from f?

src/LinearSolvers/Multigrid.h
```diff
@@ -0,0 +1,669 @@
...
+    protected:
+        std::vector<multigrid::Level<Field>> L_;
+        OperatorF op_;
+        unsigned nu1_, nu2_;
+        double omega_;
+        unsigned min_cells_per_rank_per_dim_;
+        bool communication_;
+        bool is_all_periodic_ = false;
+
+        // --- DEBUGGING - To be deleted ---
```
s-mayani: remove "to be deleted"

src/LinearSolvers/Multigrid.h
```diff
@@ -0,0 +1,669 @@
...
+        //     smooth_jacobi(level, nu1_);
+        //     dbg("after-presmooth");
+        //     restrict_average(level);
+        //     dbg("after-restrict");
+        //     vcycle(level + 1);
+        //     prolong_add(level);
+        //     dbg("after-prolong");
+        //     smooth_jacobi(level, nu2_);
+        //     dbg("after-postsmooth");
+        // }
```
s-mayani: I think I would remove this

src/PoissonSolvers/PoissonCG.h
```diff
@@ -80,13 +81,25 @@ namespace ippl {
...
                 Inform warn("PoissonCG");
-                // After reading in preconditioner parameters, if they are invalid, 
+                // After reading in preconditioner parameters, if they are invalid,
                 // the user is warned that the parameter is invalid, and a default
                 // parameter is used.
                 preconditioner_validation::sanitizeParams(preconditioner_type, warn, level, degree,
-                                                         richardson_iterations, inner, outer, omega,
-                                                         &communication);
+                                                          richardson_iterations, inner, outer,
+                                                          omega, &communication);
```
s-mayani: shouldn't the multigrid parameters also be included in the ::sanitzeParams?

s-mayani (commented):
Left some comments

src/LinearSolvers/Multigrid.h
NickvonWit: changed as much as possible. 
it is not possible [here](https://github.com/NickvonWit/ippl/blob/MG-PCG/src/LinearSolvers/Multigrid.h#L239-L239), because the communicator does not support it.

s-mayani: okay that's fine

src/LinearSolvers/Multigrid.h
NickvonWit: i just checked the code. the comment was hallucinated. deep_copy or just assigment does the same thing and keeps the boundary conditions of the result field intact, but i think this way is faster. but not sure.

src/LinearSolvers/Multigrid.h
NickvonWit: I added the removals everywhere to try to fix the convergence of the preconditioner. but it seemed to have no effect on the behaviour. I dont feel like my knowledge is enough to say whats correct.

s-mayani: @srikrrish ?

srikrrish: The volume average need to be subtracted from right hand side (I guess `f` here is rhs right?) only for consistency (as the integral of final rhs has to be zero for a periodic Poisson problem https://scicomp.stackexchange.com/questions/43026/solving-poissons-equation-with-periodic-boundary-conditions). So this seems correct to me. But @s-mayani your question is also that is it somehow done twice? I don't immediately see it but may be I am not seeing it carefully.

Having said this the final solution you obtain is unique only up to a constant (please see the above post itself), so it might be a good idea to remove the volume average from the solution also to check for the convergence. But I think you obtain correct convergence order and just only large iteration counts at certain node counts so I am not sure if this statement helps there. When you calculate electric field from potential for the mini-apps like Landau damping the constant is usually not an issue as it gets cancelled in the finite difference subtraction.

src/LinearSolvers/Multigrid.h
srikrrish: This need to be done for `f` only and doing for solution `u` is somewhat optional please see the detailed comments below.

src/LinearSolvers/Multigrid.h
srikrrish: I just checked above and there it is done for the finest level (level 0) and here for all coarse levels. So this all seems fine to me and it should be there. Please put it back in case you have removed it .

s-mayani: I think my question was more we seem to be doing this removal of the volume average at every level, but to my understanding it should only be done once at the finest level, no?

s-mayani: ah I see your above comment answers my question if it needs to be done for all levels and for both f and u for consistency

srikrrish: > I think my question was more we seem to be doing this removal of the volume average at every level, but to my understanding it should only be done once at the finest level, no?

Since we rediscretize the equation at the coarse levels for the matrix-free multigrid and not doing Galerkin coarse grid construction (for matrix based MG where the coarse grid equation is RAP = Rf where R and P are restriction and prolongation operators and A, f are the fine level matrix and rhs) the subtraction need to be done for all levels as otherwise consistency won't be automatically satisfied..

s-mayani: Okay I see, then I resolve these comments.

s-mayani (approved):

s-mayani:

srikrrish (approved):

aaadelmann:

aaadelmann:

aaadelmann:

PR #561, opened 2026-07-06
Fix opalx release assignment segfault

aliemen:
OPALX-project/OPALX#426 documents a `BareField` assignment bug in IPPL. When compiling OPALX in Release mode on GPU, we get a reliable seg-fault in the field assignment. This bug seems to be 2-sided.

## 1. Implicit `this` capture of `BareField`

The first bug was partially fixed before in #465 by replacing the whole kernel by a simple `Kokkos::deep_copy`. Since it used a `KOKKOS_CLASS_LAMBDA` and since it passed `dview_m` (which would be resolved to `this->dview_m`), the whole object is copied to device. However, `BareField` contains raw pointers like `Layout_t* layout_m = nullptr;`, which are not GPU compatible. This can become undefined behavior, instead we could capture the view directly instead of the whole `BareField` class. Meaning, the following code:
```cpp
template <typename T, unsigned Dim, class... ViewArgs>
BareField<T, Dim, ViewArgs...>& BareField<T, Dim, ViewArgs...>::operator=(T x) {
	using index_array_type = typename RangePolicy<Dim, execution_space>::index_array_type;
	ippl::parallel_for(
		"BareField::operator=(T)", getRangePolicy(dview_m),
		KOKKOS_CLASS_LAMBDA(const index_array_type& args) { apply(dview_m, args) = x; });
	return *this;
}
```
...can be fixed by adding a `auto view = dview_m;` before the kernel and by using a `KOKKOS_LAMBDA` instead. In this case, `Kokkos::deep_copy` was of course way easier. However, this fix is required for the second (more severe) bug.

## 2. Undefined behavior/alignment issue in expression templates

`BareField` inherits from `ippl::detail::Expression`. Then, the assignment operator works with an `Expression` as an argument and uses `CaptureExpression` to access the old object that's part of the expression:
```cpp
template <typename T, unsigned Dim, class... ViewArgs>
template <typename E, size_t N>
BareField<T, Dim, ViewArgs...>& BareField<T, Dim, ViewArgs...>::operator=(
	const detail::Expression<E, N>& expr) {
	using capture_type     = detail::CapturedExpression<E, N>;
	capture_type expr_     = reinterpret_cast<const capture_type&>(expr);
	using index_array_type = typename RangePolicy<Dim, execution_space>::index_array_type;
	ippl::parallel_for(
		"BareField::operator=(const Expression&)", getRangePolicy(dview_m, nghost_m),
		KOKKOS_CLASS_LAMBDA(const index_array_type& args) {
			apply(dview_m, args) = apply(expr_, args);
		});
	return *this;
}
```
This shows how `CapturedExpression` expression is used. However, internally it looks like this:
```cpp
template <typename E, size_t N = sizeof(E)>
struct CapturedExpression {
	constexpr static unsigned dim = E::dim;

	template <typename... Args>
	KOKKOS_INLINE_FUNCTION auto operator()(Args... args) const {
		static_assert(sizeof...(Args) == dim || dim == 0);
		return reinterpret_cast<const E&>(*this)(args...);
	}

	char buffer[N];
};
```
The problem now is how the data memory is treated. First assume the following simple expression: `rho1 = rho2`. Then, `rho2` is byte-interpreted as `char[N]`. A `char` has alignment 1 (see e.g. [here](https://en.cppreference.com/c/language/object#:~:text=Alignment,-Every)), but "`Args... args`" (in this case `rho2`) might have a different alignment. It contains e.g. an `int nghost_m`, which has alignment 4. Then, `reinterpret_cast` treats the `rho2` memory as a simple `char` array - which in itself is OK. However, in order to retrieve the actual object again, `BareField::operator=(` has to do `reinterpret_cast<const capture_type&>(expr);`, which **does not produce any CPU instructions**. See [here](https://en.cppreference.com/cpp/language/reinterpret_cast):

> It is primarily a compile-time directive which instructs the compiler to treat expression as if it had the type target-type.

What happens if you interpret an object in memory that has alignment 1 as another object that has alignment 4? This is undefined behavior. The solution is to simply cast the expression explicitly:
```cpp
template <typename T, unsigned Dim, class... ViewArgs>
template <typename E, size_t N>
BareField<T, Dim, ViewArgs...>& BareField<T, Dim, ViewArgs...>::operator=(
	const detail::Expression<E, N>& expr) {
	const E expr_          = static_cast<const E&>(expr);
	auto view              = dview_m;
	using index_array_type = typename RangePolicy<Dim, execution_space>::index_array_type;
	ippl::parallel_for(
		"BareField::operator=(const Expression&)", getRangePolicy(view, nghost_m),
		KOKKOS_LAMBDA(const index_array_type& args) {
			apply(view, args) = apply(expr_, args);
		});
	return *this;
}
```
Note how now `static_cast<const E&>(expr);` will (unlike `reinterpret_cast`) actually call a constructor and make sure that whatever `char[N]` is will be instantiated as a correct "`E`" type. 

## Why? Is there a performance trade off?

The main difference would be whatever the constructor does. The cast itself as well as the memory footprint/copy volume is essentially the same. We would only see a performance impact if the copy constructor was big (e.g. additional checks that go beyond simple data copies). As far as I can see, we don't have this kind of copy constructor. However, I would maybe still recommend testing at least some smaller scalings (e.g. on Daint) first. 

---

Note that I also fixed the same undefined behavior (even though we never saw it there) in `ParticleAttrib.hpp`. With these fixes, I didn't get any field assignment bugs in OPALX. Only the `accumulateHalo` seg-fault remains - meaning that this one is probably unrelated. After this PR was merged, I'll close the OPALX issue and open another one for the `accumulateHalo` bug.

### DIFF
--- a/src/Communicate/BufferHandler.h
+++ b/src/Communicate/BufferHandler.h
@@ -154,8 +154,8 @@ namespace ippl {
         buffer_type reallocateLargestFreeBuffer(size_type requiredSize);
         buffer_type allocateNewBuffer(size_type requiredSize);
 
-        size_type usedSize_m = 0;  ///< Total size of all allocated buffers
-        size_type freeSize_m = 0;  ///< Total size of all free buffers
+        size_type usedSize_m;  ///< Total size of all allocated buffers
+        size_type freeSize_m;  ///< Total size of all free buffers
 
     protected:
         buffer_set_type used_buffers{
--- a/src/Communicate/BufferHandler.h
+++ b/src/Communicate/BufferHandler.h
@@ -154,8 +154,8 @@ namespace ippl {
         buffer_type reallocateLargestFreeBuffer(size_type requiredSize);
         buffer_type allocateNewBuffer(size_type requiredSize);
 
-        size_type usedSize_m;  ///< Total size of all allocated buffers
-        size_type freeSize_m;  ///< Total size of all free buffers
+        size_type usedSize_m = 0;  ///< Total size of all allocated buffers
+        size_type freeSize_m = 0;  ///< Total size of all free buffers
 
     protected:
         buffer_set_type used_buffers{
--- a/src/Communicate/Communicator.h
+++ b/src/Communicate/Communicator.h
@@ -16,7 +16,6 @@
 ////////////////////////////////////////////////
 // For message size check; see below
 #include <climits>
-#include <cstring>
 #include <cstdlib>
 
 #include "Utility/TypeUtils.h"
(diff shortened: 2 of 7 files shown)


aliemen:
Use [this diff](https://github.com/IPPL-framework/ippl/pull/561/changes/BASE..e63d6fc4a42c68f94e7b6eefdcf9a4d818025924) for the actual changes. `clang-format` added a lot of diff in `ParticleAttrib`...

aliemen:
I found another occurence of the second bug mentioned above inside `FEMVector.h`. I applied the same fix there, but had to clang-format. [Here](https://github.com/IPPL-framework/ippl/pull/561/changes/940f44174afd0626b5de78d43bcd62e60f303cc8) is the actual diff for the `FEMVector` fix.

aliemen:
@aaadelmann all checks now passed, including [Jülich](https://ippl-bc4558.pages.jsc.fz-juelich.de/477172/allure-report/index.html). I'm also quite certain that I found the `accumulateHalo` bug this morning, I will make a second PR for that.

aaadelmann (approved):

PR #570, opened 2026-07-20
Pr501 fft final

aaadelmann:
```md
## Summary

This PR ports the FFT backend refactor from the larger PR501 branch onto current `master`, keeping the scope limited and reviewable.

The main change is to split the former monolithic FFT implementation into smaller backend and transform modules while preserving the existing public FFT usage for the current HeFFTe-backed transforms.

## What Is Included

### FFT Core Refactor

The FFT implementation is now organized into separate headers:

- `src/FFT/Traits.h`
- `src/FFT/Backend/Backend.h`
- `src/FFT/Backend/Heffte.h`
- `src/FFT/Transform/Common.h`
- `src/FFT/Transform/CC.h`
- `src/FFT/Transform/RC.h`
- `src/FFT/Transform/Trig.h`
- `src/FFT/Transform/Transform.h`

`src/FFT/FFT.h` is now the aggregate include for the refactored FFT module.

The old monolithic implementation header:

- `src/FFT/FFT.hpp`

was removed.

### Pruned Transforms

This PR also adds the pruned transform specializations:

- `src/FFT/Transform/PrunedCC.h`
- `src/FFT/Transform/PrunedRC.h`

These are included through the FFT transform aggregate header.

### Optional cuFFTMp Backend

This PR adds the multi-node NVIDIA cuFFTMp backend as an explicitly opt-in backend:

- `src/FFT/Backend/CuFFTMp.h`

A new CMake option was added:

```cmake
-DIPPL_ENABLE_CUFFTMP=ON
```

The option is guarded so that cuFFTMp can only be enabled when FFT and CUDA are both enabled:

- `IPPL_ENABLE_CUFFTMP=ON` requires `IPPL_ENABLE_FFT=ON`
- `IPPL_ENABLE_CUFFTMP=ON` requires `CUDA` in `IPPL_PLATFORMS`

The backend header is included only when both conditions are active:

```cpp
#if defined(IPPL_ENABLE_CUFFTMP) && defined(KOKKOS_ENABLE_CUDA)
#include "FFT/Backend/CuFFTMp.h"
#endif
```

Plain single-node `CuFFT.h` support is intentionally not included in this PR.

## What Is Intentionally Not Included

The original `pr501-fft` branch also contained changes to `FFTOpenPoissonSolver` and solver tests. Those were reviewed but not ported.

Reason: current `master` already contains newer Poisson solver functionality, including integrated and shifted Green-function support. The corresponding hunks from `pr501-fft` would remove that current coverage and regress master.

Specifically, this PR does **not** remove:

- `GreenFunction::{STANDARD, INTEGRATED}`
- integrated Green-function support
- shifted integrated Green-function support
- `TestGaussianIntegrated`
- `TestIntegratedGreensFunction`
- current CG, preconditioner, and multigrid tests

## Validation

Configured and built an OpenMP FFT/unit-test build:

```bash
cmake -S . -B build-pr501-fft-final-openmp \
  -DIPPL_PLATFORMS=OPENMP \
  -DIPPL_ENABLE_FFT=ON \
  -DIPPL_ENABLE_UNIT_TESTS=ON \
  -DIPPL_ENABLE_TESTS=ON

cmake --build build-pr501-fft-final-openmp -j 8
ctest --test-dir build-pr501-fft-final-openmp --output-on-failure
```

Result:

```text
100% tests passed, 0 tests failed out of 70
```

Configured and built a solver-enabled OpenMP build:

```bash
cmake -S . -B build-pr501-fft-final-solvers-openmp \
  -DIPPL_PLATFORMS=OPENMP \
  -DIPPL_ENABLE_FFT=ON \
  -DIPPL_ENABLE_SOLVERS=ON \
  -DIPPL_ENABLE_UNIT_TESTS=ON \
  -DIPPL_ENABLE_TESTS=ON

cmake --build build-pr501-fft-final-solvers-openmp -j 8
ctest --test-dir build-pr501-fft-final-solvers-openmp --output-on-failure
```

Result:

```text
100% tests passed, 0 tests failed out of 85
```

Also verified the cuFFTMp CMake guard:

```bash
-DIPPL_PLATFORMS=OPENMP -DIPPL_ENABLE_FFT=ON -DIPPL_ENABLE_CUFFTMP=ON
```

fails clearly with:

```text
IPPL_ENABLE_CUFFTMP requires CUDA in IPPL_PLATFORMS
```
## Scaling 
image

## Notes

CUDA/cuFFTMp backend code is included behind compile-time guards and was not fully exercised locally because the local build is OpenMP-only. The default path remains HeFFTe, and non-CUDA builds do not include the cuFFTMp backend header.
```

### CODE REVIEW

File: src/FFT/Traits.h [hunk 1]
```diff
@@ -0,0 +1,323 @@
...
+        template <>
+        struct is_available<HeffteGPU> : std::true_type {};
+#endif
+
+#if defined(IPPL_ENABLE_CUFFTMP) && defined(KOKKOS_ENABLE_CUDA)
+        template <>
+        struct is_available<CuFFTMp> : std::true_type {};
+#endif
+
+#ifdef ENABLE_FINUFFT
```
srikrrish, line 138: FINUFFT related enables are not there in Dependencies.cmake. Since it is related to NUFFT will it come in the next PR possibly? If yes then it is better to remove them in this one?

src/FFT/Traits.h
```diff
@@ -0,0 +1,323 @@
...
+        template <>
+        struct is_available<Finufft> : std::true_type {};
+#endif
+
+#if defined(ENABLE_FINUFFT) && defined(ENABLE_GPU_NUFFT)
```
srikrrish, line 143: Same comment here with respect to FINUFFT

srikrrish (changes requested):
I just noticed that apart from the PCG PR Paul is not mentioned as author in the previous PR (HOSG final) as well as this one. This needs to be fixed. Since HOSG final is already merged it may be difficult to correct it there and not sure (sorry should have checked more carefully). but hopefully at least in this PR and in the future ones it can be corrected.

src/FFT/Traits.h
aaadelmann: FINUFFT comes in the next PR, and this one is just an intermediate so I guess this is ok

aaadelmann:
Will add Paul to this one and see if we can add him to the other as well.

srikrrish (approved):
It was decided that Paul will be added as a co-author at the end of all the PRs from 501 by selecting the individual commits. Hence, I am approving now.

PR #578, opened 2026-07-27
P3M Part 1: "Fix" periodic P3M algorithm

### LINKED ISSUE #575: P3M long range Green's function and PP interaction can be improved
At the moment, the P3M PP interaction does not regularize and the Green's function for the long range interaction can be calculated directly in Fourier space.

For a more detailed explanation, please refer to [my OPALX comment](https://github.com/OPALX-project/OPALX/issues/489#issuecomment-5057540920).

With this Issue/PR I will change the Green's function (which reduces the number of FFTs) and implement a "naive" regularization.

aliemen:
closes #575 

This PR fixes the P3M long range interaction by directly computing the Green's function in Fourier space. This saves on FFT call and improves accuracy. Additionally, it introduces a simple uniform regularization (easiest solution, will probably be refined in the future). Already existing tests were improved such that they now don't "trivially pass". 

Since it is relevant, I'll post my comment from the OPALX-project/OPALX#489 issue again. In the near future, I plan to generalize the P3M implementation also to open BCs (separate PR). 

---

After adapting the Green's function in IPPL and adding regularization to the interaction term, we get:

Image

The input uses a Gaussian $100\mathrm{pC}$ bunch with $\sigma = 1\mathrm{mm}$ and $p_0 = 2\cdot 10^{-3} \cdot m_\mathrm{e}$ (non relativistic). All tests use `8ppc`.

## New Green's function

Before, $G$ was constructed in real space and then Fourier-transformed. This introduces some error and instead we can construct the exact periodic Ewald operator in Fourier space. Before:
```math
r_j^2 = \sum_d h_d^2 \,\mathrm{min} (j_d, N_d - j_d )^2
```
with real space Green's function $g_j \propto \frac{\mathrm{erf}(\alpha r_j)}{r_j}$, and then the transformed $\tilde{g}_k = \mathrm{FFT}[g_j]$. The transform on one "minimum-image copy" is not the complete periodic Ewald sum and introduced error. Instead, we can construct the calculate the Fourier transformed Green's function directly:
```math
\tilde{g}_k \propto \frac{.\mathrm{exp}[ -k^2 / (4 \alpha^2) ] }{k^2}, \qquad k \neq 0 .
```
This even saves one FFT and makes the algorithm faster. 

## New regularization

The current IPPL algorithm computes the plain particle particle interaction for every combination for all $r < r_\mathrm{cut}$:
```math
\mathbf{E}_\mathrm{PP} (\mathbf{r}) \propto - \nabla \left[ \frac{\mathrm{erfc}(\alpha r)}{r} \right] .
```
Here, $r = \left| \mathbf{R}_i - \mathbf{R}_j \right|$ is not regularized, so the division by $r$ can blow up the field. For P3M (albeit this comes from cosmology, where we have actual "homogeneous mass distributions") a uniform sphere Kernel is often used, i.e.:
```math
\mathbf K_\epsilon(\mathbf r)=
\begin{cases}
\mathbf r/\epsilon^3, & r<\epsilon,\\[1mm]
\mathbf r/r^3, & r\geq\epsilon.
\end{cases}
```
From this we can also calculate the full regularized potential:
```math
\phi_{\mathrm{reg}}(r) \propto
\begin{cases}
\dfrac{3}{2 \varepsilon}
-\dfrac{r^2}{2 \varepsilon^3},
& r < \varepsilon, \\[6pt]
\dfrac{1}{r},
& r \geq \varepsilon.
\end{cases}
```
This corresponds to a Coulomb interaction for $r > varepsilon$ and a uniformly charged sphere for `r < \varepsilon` and is continuous at `r = \varepsilon`, see e.g. [Figure 1 in this article](https://doi.org/10.1006/jcph.1997.5723).

Together, with the subtraction of the long range interaction and with the regularization, the complete regularized interaction for $r < \varepsilon$ onto particle $i$ becomes:
```math
\Delta\mathbf E^{\mathrm{PP,reg}}_{i\leftarrow j}
=
k_eq_j\mathbf r_{ij}
\left[
\frac{1}{\varepsilon^3}
-
\frac{\mathrm{erf}(\alpha r)}{r^3}
+
\frac{2\alpha}{\sqrt{\pi}}
\frac{e^{-\alpha^2r^2}}{r^2}
\right].
```
It turns out that all we have to do is replace $r$ by $\varepsilon$ in the short range interaction.


src/Interaction/TruncatedGreenParticleInteraction.hpp
```diff
@@ -1,26 +1,35 @@
 //
 // Class TruncatedGreenParticleInteraction
-//   This class implements the short range interaction part of the green function splitting obtained
-//   by taking the gradient of [forceConstant * (1 - erf(alpha * r)) / r]. The long range part is
-//   handled by FFTTruncatedGreenPeriodicSolver.
+//   This class implements the short-range part of the Ewald Green function split by evaluating the
```
aaadelmann, line 3: Only add header in .h file not the same again in .hpp

test/particle/TestTruncatedGreenParticleInteraction.cpp
```diff
@@ -0,0 +1,125 @@
+#include "Ippl.h"
```
aaadelmann, line 13: Describe what is being tested here

aaadelmann (changes requested):
We would also need a unit_test for this

aliemen:
@aaadelmann Will fix the two requested changes now.

Regarding the unit test. I followed what e.g. the open Poisson solver does: at the moment there are two integration tests `test/particle/TestTruncatedGreenParticleInteraction.cpp` and `test/solver/TestFFTTruncatedGreenPeriodicPoissonSolver.cpp`. Those test the solver itself, but aren't unit tests, since they might be too complex for a unit test. Should I just move them in the unit test folder regardless (both integration and unit tests are run by the CI)?

aaadelmann:
ok then lets stay with the integration tests, no unit tests

src/Interaction/TruncatedGreenParticleInteraction.hpp
aliemen: Done.

test/particle/TestTruncatedGreenParticleInteraction.cpp
aliemen: Done.

aliemen:
Since the CI is always failing, I also pushed a fix for `LandauDampingCorrectnessValidation`: 

The problem is that the executable tries to access the `.csv` files that still lives inside the source code directory. However, once the build is done and the execution step starts, there is no guarantee that the source dir persists. If it does, the CI passes. If it was (automatically?) cleaned up, it fails... The easy fix is to just add something like the following to the LandauDamping cmake:
```cmake
configure_file(
  "${CMAKE_CURRENT_SOURCE_DIR}/FieldLandau_valid_result.csv"
  "${landau_reference_csv}"
  COPYONLY)
```
This will make sure that the reference `.csv` is automatically copied to the build directory and the file path issue should be fixed. Since the `.csv` file is just a few lines, that's not a problem.

aaadelmann (approved):

PR #581, opened 2026-07-31
Improve Landau test setup/cleanup and validation

biddisco:
Cleanup Landau validation
improvement is that pre-test - we clean the output, then run the test, then validate the test. This prevents a collision if the test is run twice causing the previous results to contaminate the new results.

### DIFF
--- a/demos/alpine/CMakeLists.txt
+++ b/demos/alpine/CMakeLists.txt
@@ -5,7 +5,6 @@
 # -----------------------------------------------------------------------------
 message(STATUS "Configuring demos/alpine/")
 
-
 add_subdirectory(ExamplesWithoutPicManager)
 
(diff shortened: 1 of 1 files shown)


demos/alpine/CMakeLists.txt
```diff
@@ -18,13 +17,28 @@ endfunction()
 
 if(IPPL_ENABLE_TESTS)
   # cmake-format: off
-  # Landau will write a CSV file to the data directory that we will validate later
```
aaadelmann, line 21: this comment should stay

aaadelmann: ok got it ...

aaadelmann (approved):

PR #595, opened 2026-08-17
Reduce size of binary build artefacts to help CI with debug builds

biddisco:
This should fix the 1GB limit for artefacts. 
Flyby changes are renaming the pipelines to use cuda13 instead of 12 and also increasing the time limit from 15 to 30 mins


biddisco:
cscs-ci run cscs-ci-gh200

aaadelmann (approved):

Issue #600, opened 2026-09-08
Enable Kokkos Kernels with explicit host and GPU dependencies


aaadelmann:
### Summary

Enable optional Kokkos Kernels support in IPPL and add a linear-map eigenanalysis unit test inspired by OPALX's `unit_tests/Algorithms/TestLinearMapEigenAnalysis.cpp`.

## Dependency requirements

Kokkos does not automatically install Kokkos Kernels or external numerical libraries.

In Kokkos Kernels 5.2.0, `SerialEigendecomposition` requires host LAPACKE or MKL. Its native/device implementation is unfinished. CUDA and HIP builds therefore still need a host provider for this eigenanalysis.

Dependencies by platform:

- Serial/OpenMP: host LAPACKE with LAPACK/BLAS, or MKL. Including MAC Os
- CUDA: the same host dependency, plus cuBLAS and cuSOLVER for supported GPU operations.
- HIP: the same host dependency, plus rocBLAS and rocSOLVER for supported GPU operations.

Enabling GPU libraries does not move this eigenanalysis onto the GPU.

## CMake tasks

- [ ] Add an `IPPL_ENABLE_KOKKOS_KERNELS` option. 
- [ ] Default IPPL_ENABLE_KOKKOS_KERNELS is enabled
- [ ] Support installed and FetchContent-built Kokkos Kernels, using a compatible version and the same Kokkos instance as IPPL.
- [ ] Configure host LAPACKE or MKL independently of the GPU backend.
- [ ] Select the appropriate CUDA/HIP libraries for supported GPU operations.
- [ ] Keep installation paths in site presets or toolchain files.
- [ ] Verify matching host headers, libraries, and LP64/ILP64 integer interfaces.
- [ ] Validate required eigenanalysis support with a compile-and-link probe.
- [ ] Report a clear configuration error when eigenanalysis is requested but unavailable.
- [ ] Propagate dependencies through CMake targets and IPPL's exported package.
- [ ] Document dependency installation, configuration options, and the host-only eigenanalysis limitation.

Configuration options cannot change the capabilities of an already installed Kokkos Kernels package. Validate the selected package rather than assuming it provides the requested backend.

## Unit test

Use small 4x4 maps with analytically known spectra, inspired by the OPALX tests.

- [ ] Stable rotations and conjugate eigenvalue pairs.
- [ ] Coupled stable modes.
- [ ] Real and complex instabilities.
- [ ] Near-integer and neutral modes.
[...]

PR #601, opened 2026-09-08
Enable Kokkos Kernels with host and GPU dependency configuration

aaadelmann:
Enable Kokkos Kernels by default, supporting installed packages and FetchContent builds. Configure host LAPACKE/MKL independently of CUDA/HIP libraries and validate provider capabilities, linkage, and integer ABI.

Add OPALX-inspired linear-map tests covering analytic eigenvalues and eigenvector residuals. Use Kernels GEMM and direct host LAPACKE because the upstream 5.2.0 eigenanalysis header fails with GCC 15 and Clang 21.

Fix build-tree/install dependency propagation, add a standalone consumer test, document configuration, and forward dependency options through CI.

Validated Serial and OpenMP builds, GCC and Clang, 1/2/4 MPI ranks, disabled configurations, and build-tree/install consumers. CUDA, HIP, MKL, and a working ILP64 provider remain unverified.

Refs #600


aaadelmann:

cmake/BuildSystem.md
```diff
@@ -172,3 +172,129 @@ target_link_libraries(app PRIVATE ippl::ippl)
 ```
 
 
+
+## Kokkos Kernels and host eigenanalysis
+
+`IPPL_ENABLE_KOKKOS_KERNELS` defaults to `ON`. IPPL first finds an installed
```
srikrrish: All our enables are by default `off` if I am not wrong as we use the opt-in model to keep it light-weight and the user selects what they may need. Should we do that for this also to be consistent?

cmake/BuildSystem.md
```diff
@@ -172,3 +172,129 @@ target_link_libraries(app PRIVATE ippl::ippl)
...
+  resolved through `LAPACKE_LIBRARY_DIRS`; absolute paths are supported.
+- `IPPL_KOKKOS_KERNELS_HOST=MKL`: provide an MKL CMake package through
+  `CMAKE_PREFIX_PATH`/`MKL_DIR` (modern oneMKL).
+- `IPPL_KOKKOS_KERNELS_HOST=NONE`: use portable/GPU kernels without requesting
+  host eigenanalysis; only the GEMM regression is registered in the test binary.
+- `IPPL_ENABLE_KOKKOS_KERNELS=OFF`: omit the dependency and its tests entirely.
+
+Kokkos does not install these external libraries. If host LAPACKE is not found,
+IPPL now downloads reference LAPACK 3.12.1 (SHA256-verified) and builds its BLAS,
+LAPACK and LAPACKE libraries. The fallback requires host C and Fortran compilers;
```
srikrrish, line 235: It also may make sense with this option. Because if it is on by default and there is no LAPACKE then it fall back to this default option and then unnecessarily needs C, Fortran compilers etc. right?

cmake/BuildSystem.md
```diff
@@ -172,3 +172,129 @@ target_link_libraries(app PRIVATE ippl::ippl)
...
+set `CMAKE_Fortran_COMPILER` or `FC` if automatic discovery cannot find gfortran.
+The host C compiler is selected independently via `CC` or the fallback toolchain.
+It supports LP64 and ILP64 through `IPPL_LAPACK_INTEGER_BYTES` and builds static
```
srikrrish, line 238: Do we need to expose these many IPPL flags?

cmake/BuildSystem.md
```diff
@@ -172,3 +172,129 @@ target_link_libraries(app PRIVATE ippl::ippl)
...
+position-independent libraries for use by shared or static IPPL.
+
+The first configure builds this dependency in an isolated host project under
+`_deps`, before the existing compile/link/runtime probes run. This can take a few
+minutes; subsequent configures reuse the build. CUDA/HIP compiler launchers and
+IPPL's directory flags are not applied to the host project. Its logs are in
+`_deps/ippl-host-lapack-<integer-bytes>-build/{configure,build,install}.log`.
+`IPPL_LAPACKE_BUILD_JOBS` controls build parallelism (default 4).
```
srikrrish, line 246: Same here.. it seems to be lot of flags (which are already quite many currently)

cmake/BuildSystem.md
```diff
@@ -172,3 +172,129 @@ target_link_libraries(app PRIVATE ippl::ippl)
...
+For offline builds, point `FETCHCONTENT_SOURCE_DIR_IPPL_REFERENCE_LAPACK` at an
+already unpacked reference LAPACK 3.12.1 source tree.
+
+Set `IPPL_FETCH_LAPACKE=OFF` to require an installed provider. An explicit
+`LAPACKE_LIBRARIES` remains authoritative: missing headers or broken linkage
+produce errors instead of silently substituting a different provider. A Kernels
+package already built without its LAPACKE TPL still needs rebuilding; fetching
+LAPACKE cannot enable features in an installed Kernels library.
+
+For a cross build, supply `IPPL_LAPACKE_TOOLCHAIN_FILE` with a C/Fortran toolchain
```
srikrrish, line 256: Here..

cmake/BuildSystem.md
```diff
@@ -172,3 +172,129 @@ target_link_libraries(app PRIVATE ippl::ippl)
...
+The first configure builds this dependency in an isolated host project under
+`_deps`, before the existing compile/link/runtime probes run. This can take a few
+minutes; subsequent configures reuse the build. CUDA/HIP compiler launchers and
+IPPL's directory flags are not applied to the host project. Its logs are in
+`_deps/ippl-host-lapack-<integer-bytes>-build/{configure,build,install}.log`.
+`IPPL_LAPACKE_BUILD_JOBS` controls build parallelism (default 4).
+For offline builds, point `FETCHCONTENT_SOURCE_DIR_IPPL_REFERENCE_LAPACK` at an
+already unpacked reference LAPACK 3.12.1 source tree.
+
+Set `IPPL_FETCH_LAPACKE=OFF` to require an installed provider. An explicit
```
srikrrish, line 250: Here..

srikrrish (commented):
Actually, we need this only for very small matrices (like 6x6 etc.) if I am not wrong right? Isn't it a bit overkill to expose these many options? I am also worried about the dependency chain but since we need it and I currently don't see a work around at least to be consistent with rest of IPPL enables it can be OFF by default?

cmake/BuildSystem.md
aaadelmann: will set by default off

cmake/BuildSystem.md
aaadelmann: ok default off

cmake/BuildSystem.md
aaadelmann: ok agree and yes a few flags but the are all preset.

cmake/BuildSystem.md
aaadelmann: all is per default off. manual updated

cmake/BuildSystem.md
aaadelmann: yes I think we do but they all have default values

srikrrish (approved):

aaadelmann:
cscs-ci run cscs-ci-openmp

PR #603, opened 2026-09-17
Resolve include loop in `ParallelDispatch.h` and `datatypes.h`

### LINKED ISSUE #602: Circular include in `ParallelDispatch.h` can cause include order compiler failure in OPALX
On current IPPL master, including `Manager/datatypes.h` without first including `Ippl.h` fails with errors such as:

```text
HaloCells.hpp:258: error: no template named 'RangePolicy'
HaloCells.hpp:262: error: no member named 'parallel_for' in namespace 'ippl'
```

`ParallelDispatch.h` includes `Ippl.h`, while `Ippl.h` includes `ParallelDispatch.h` and then `IpplCore.h`. This cycle allows `HaloCells.hpp` to be parsed before `ippl::RangePolicy` and `ippl::parallel_for` are declared. Headers should compile independently of whether `Ippl.h` was included first. Removing the `Ippl.h` dependency from `ParallelDispatch.h`, or otherwise breaking the cycle, should resolve the issue. 

This include chain lead to a compiler error in OPALX. In can also be circumvented in OPALX by including `Ippl.h` first, but the "cleaner" fix would be here in IPPL.

aliemen:
closes #602 

This PR removes `#include "Ippl.h"` from `ParallelDispatch.h` and explicitly adds it to `datatypes.h`. This breaks the include order explained in the attached issue. The "`threadSafe`" variable is now queried directly without `Env`. The rest should be clang-format noise.


aaadelmann (approved):

aliemen:
cscs-ci run cscs-ci-mi300

aliemen:
Tested OPALX-project/OPALX#557 against this branch on Daint: everything (unit and regression tests) worked, even the temporarily deactivated 2.5D unit tests.

